PR #313, opened 2022-10-28
Switch to a Mariner-based and multiarch container image

gracewehner:
Latest version for testing: **5.4.0-grace-marinerv2-11-03-2022-03f37bfe**

Previous Mariner PR: https://github.com/Azure/prometheus-collector/pull/211

Mariner:
- Dockerfile changes to use mariner instead of ubuntu base image. This now uses the Docker multi-stage build to use the package manager with the regular mariner base image to install our dependencies, and then copy them over into the distroless container
- Changes to setup.sh and main.sh due to the package manager being tdnf and so the install commands are different. The commands/parameters to run some executables are also different
- Extra files needed for crontab and logrotate that are included by default for ubuntu but not mariner
- Extra files needed for the mariner-extras repo to install ME and mdsd. Other mariner repos are included on the container by default
- Add telemetry for Mariner host nodes
- New internal doc to explain more about mariner development and the differences between it and ubuntu

Multiarch:
- Build otelcollector, prom-config-validator, and fluent-bit plugin with the Mariner-based golang container so we can build for both AMD64 and ARM64 without needing an ARM64 environment
- Change docker build commands for multiarch and adding TARGETOS, TARGETARCH arguments to dockerfile to designate when to split off building for either AMD64 or ARM64 when installing packages
- Add telemetry for ARM64 nodes
- New internal doc to explain more about building multiarch images

Other:
- Tweak ME config settings to fix ContentLengthLimitExceeded GIG issue. This is also seen in our current prod ubuntu-based image
- Refactor our core capacity count telemetry so that less code is repeated

Image size comparisons:

- ubuntu-based image (our current prod image size): 882.36MB
- non-distroless mariner-based multiarch image: 1.3GB
- distroless mariner-based multiarch image: 597.54 MB


otelcollector/metricextension/me.config
```diff
@@ -11,9 +11,9 @@
       "honorResourceAttributes":true,
       "maxReceiveMessageSizeMBytes": 12
    },
-   "publicationIntervalInSec":30,
+   "publicationIntervalInSec":20,
    "maxPublicationAttemptsPerMinute":12,
-   "maxPublicationPackageSizeInBytes":100000000,
+   "maxPublicationPackageSizeInBytes":2000000,
```
vishiy, line 16: any impact on the scale tests, Grace for moving to 2mb from 100mb ?

otelcollector/scripts/main.sh
```diff
@@ -286,27 +294,26 @@ echo_var "OTELCOLLECTOR_VERSION" "$OTELCOLLECTOR_VERSION"
...
-#get ruby version
-RUBY_VERSION=`ruby --version`
-echo_var "RUBY_VERSION" "$RUBY_VERSION"
-
-echo "Starting telegraf"
+echo "starting telegraf"
 if [ "$TELEMETRY_DISABLED" != "true" ]; then
-  /opt/telegraf/telegraf --config /opt/telegraf/telegraf-prometheus-collector.conf &
-  TELEGRAF_VERSION=`/opt/telegraf/telegraf --version`
+  telegraf --config /opt/telegraf/telegraf-prometheus-collector.conf &
```
vishiy, line 299: doesn't mdsd drop telegraf as well? Isn't better to point to the right telegraf's path when starting it ?

otelcollector/metricextension/me.config
gracewehner: Things look good with this combined with decreasing the publication interval from 30s to 20s as Emre recommended. For the image from this PR vs our last scale testing, the perf has improved with the same benchmarks for scraping all default targets on a large cluster: 1.7 Gi before vs 1.4 Gi now and 1.1 cores vs. 500mc now for replicaset and confirmed it's the same total amount of metrics are being sent

otelcollector/scripts/main.sh
gracewehner: Yes right now we are removing it though in setup.sh. I will add the full path just in case though, good point

otelcollector/metricextension/me.config
vishiy: thanks grace

vishiy (approved):

vishiy (approved):

PR #354, opened 2023-01-05
Upgrade dependencies

gracewehner:
- Upgrade otelcollector, ME, and mdsd
- Fix external labels bug in otelcollector
- Fix file rotation in distroless so that the correct permission and execute paths are set for cron


otelcollector/scripts/setup.sh
```diff
@@ -36,12 +36,13 @@ gem install re2
 
 echo "Installing mdsd..."
 if [ "${ARCH}" != "amd64" ]; then
-  wget https://github.com/microsoft/Docker-Provider/releases/download/mdsd-mac-official-06-13/azure-mdsd_1.19.3-build.master.428_aarch64.rpm
-  sudo tdnf install -y azure-mdsd_1.19.3-build.master.428_aarch64.rpm
+  wget https://github.com/Azure/prometheus-collector/releases/download/azure-mdsd-1.23.3/azure-mdsd_1.23.4-build.master.28_aarch64.rpm
```
vishiy, line 39: grace, mdsd is still not part of mariner packages ?

otelcollector/prometheusreceiver/config.go
```diff
@@ -18,7 +18,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
+	"net/url"
```
vishiy, line 21: grace, can you add it somewhere the version & commit we have been synced to for prom receiver ?

gracewehner: Right, thanks I updated the files we have for prom-version and otelcollector version. The prom receiver is synced to the v0.66.0 tag (same as the otelcollector version) so there are no conflicts

otelcollector/scripts/setup.sh
gracewehner: It is but the version with the fix for the extra error logging hadn't been published yet as a Mariner package at the time, so mdsd team sent over the RPMs for 1.23.4. 1.23.5 (there's no mariner package for 1.23.4) is now published, but this may reset some testing :) I'm happy to change it and re-test, or I'll add a task to switch it back over for the next release?

.trivyignore
```diff
@@ -40,3 +40,4 @@ CVE-2022-46146
 # kube-state-metrics
 # bin/node_exporter
 CVE-2022-32149
+CVE-2022-41717
```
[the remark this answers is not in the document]
gracewehner: Thanks, removed the CVE for otelcollector/promconfigvalidator. For ME and mdsd there weren't any CVEs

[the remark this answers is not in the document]
vishiy: ok thank you.

otelcollector/scripts/setup.sh
vishiy: ok we can pick up next time.

vishiy (approved):

PR #371, opened 2023-01-27
Add pod annotation based scraping as a default target

gracewehner:
Test cases:
- Enabled when no settings configmap exists
- Enabled when configmap exists but doesn't have any of the new pod annotation settings in it
- Disabled when set to false in the configmap
- Metricskeeplist custom setting works
- Custom scrape interval setting works

Will create a branch in azure-docs-pr for what should be published when these changes roll out


otelcollector/configmaps/ama-metrics-settings-configmap.yaml
```diff
@@ -30,7 +31,10 @@ data:
     nodeexporter = ""
     windowsexporter = ""
     windowskubeproxy = ""
+    podannotations = ""
     minimalingestionprofile = true
+  default-targets-namespace-keep-list: |-
+    podannotations = ""
```
vishiy, line 37: empty --  meaning all namespaces would be included ? How would one include only a few namespaces? by specifying it here ?  For our metric keep lists - the keeplist is in addition to what we collect by default.

otelcollector/configmaps/ama-metrics-settings-configmap.yaml
```diff
@@ -30,7 +31,10 @@ data:
     nodeexporter = ""
     windowsexporter = ""
     windowskubeproxy = ""
+    podannotations = ""
```
vishiy, line 34: may be we should have just 1 setting for podannotation (called podannotation namespaces -- outside targets list). If its empty pod annotation is turned off . if it is not empty , podannotation is true for the specified namespaces ? A thought, as i feel this is a bit more confusing now with the new 'namespace keep list' pivot to the configmap.

PR #376, opened 2023-02-01
fix: only import proper certificate files

bragi92:


otelcollector/build/windows/scripts/main.ps1
```diff
@@ -367,8 +367,15 @@ function Set-CertificateForME {
     Copy-Item -r /etc/config/settings/akv /opt/akv-copy
 
     Get-ChildItem "C:\etc\config\settings\akv\" |  Foreach-Object {
-        if (!($_.Name.startswith('..'))) {
-            Import-PfxCertificate -FilePath $_.FullName -CertStoreLocation Cert:\CurrentUser\My > $null
+        # check if child is a file and not a directory
```
vishiy, line 370: kaveesh - what was the issue here ? we are trying to import anything in there ? But everything in there should be certs, isnt it ?

bragi92: The below list is what we're iterating on to import. Previously it only had the PFX file and the directories so I was ignoring the directories and importing the certificate. Now it looks like two null .crt and .key files are also being mounted in the container which throw an error when I try to import them since they're null. I've update it so that I try to import any non null file within the said folder.

image

vishiy: ok. thanks. is this only an issue on windows ?

vishiy: thanks.

vishiy (approved):

PR #377, opened 2023-02-02
AKS addon HTTP Proxy Support

gracewehner:
Support using [AKS HTTP cluster proxy settings](https://learn.microsoft.com/en-us/azure/aks/http-proxy).

The following env variables are automatically passed to the pods (even through the backdoor deployment) by AKS:
- HTTP_PROXY
- http_proxy
- HTTPS_PROXY
- https_proxy
- NO_PROXY
- no_proxy

The optional trusted CA cert is added to the node by AKS in either the directory /usr/local/share/ca-certificates/ for Ubuntu or /etc/pki/ca-trust/extracted for Mairner. We need to mount these and update the trusted certs at startup. 

Most processes will automatically recognize and use the above variables automatically.
- golang automatically recognizes these so fluent-bit and telegraf don't require any changes
- ME will automatically recognize these with "proxyDefinitionMode":1 (read from env vars)
- mdsd recognizes these but sees the optional '/' character at the end as invalid, so I am removing this at startup

Added to our telemetry from telegraf the dimension `HTTP_PROXY_ENABLED` if HTTP_PROXY is set


vishiy:
Grace, please add a work item to doc this externally for next release.

vishiy (approved):

PR #379, opened 2023-02-08
Enable using workload identities for accessing key vault secrets

lnr0626:
This PR adds support for workload identities into the prometheus collector chart. When `azureKeyVault.useManagedIdentity` is `true`, users can use `azureKeyVault.clientId` to provide the client id for an identity that is configured to use workload identities.

If multiple identity properties are specified, it will use `aad-pod-identity`, then `clientId`, then `userAssignedIdentityID`, and finally the system assigned managed identity.

### DIFF
--- a/otelcollector/deploy/chart/prometheus-collector/templates/prometheus-collector-secretProviderClass.yaml
+++ b/otelcollector/deploy/chart/prometheus-collector/templates/prometheus-collector-secretProviderClass.yaml
@@ -14,7 +14,7 @@ spec:
   parameters:
     {{- if ne (index .Values.azureKeyVault "aad-pod-identity") ""}}
     usePodIdentity: "true"
-    {{- else }}
+    {{- else if ne .Values.azureKeyVault.userAssignedIdentityID "" }}
     usePodIdentity: "false"                    # [OPTIONAL] if not provided, will default to "false"
     useVMManagedIdentity: {{ .Values.azureKeyVault.useManagedIdentity | toString | quote }}           # [OPTIONAL available for version > 0.0.4] if not provided, will default to "false"
     {{- if .Values.azureKeyVault.userAssignedIdentityID }}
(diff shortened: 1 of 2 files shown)


otelcollector/deploy/chart/prometheus-collector/templates/prometheus-collector-secretProviderClass.yaml
```diff
@@ -39,5 +41,5 @@ spec:
           objectFormat: pfx                                  # [REQUIRED] DO NOT CHANGE FROM 'pfx'
           objectVersion: ""                                  # [OPTIONAL] object versions, default to latest if empty
       {{- end}}
-    tenantId: {{ required "azureKeyVault.tenantId is required" .Values.azureKeyVault.tenantId | toString | quote }}       # [CHANGE AS APPROPRIATE][REQUIRED] the tenant ID of the KeyVault specified above
+    tenantID: {{ required "azureKeyVault.tenantId is required" .Values.azureKeyVault.tenantId | toString | quote }}       # [CHANGE AS APPROPRIATE][REQUIRED] the tenant ID of the KeyVault specified above
```
vishiy, line 44: is it intentional to change `tenantId` to `tenantID` ? I see its being referenced when used as `tenantId` .

lnr0626: Yes, tenantId has been deprecated for removal in favor of tenantID inside the SPC

lnr0626:
Sorry for the delayed response, I never saw the notification that you'd reviewed this pr. 

Thanks for taking a look!

otelcollector/deploy/chart/prometheus-collector/templates/prometheus-collector-secretProviderClass.yaml
vishiy: Ok, thanks. I see the PR here -- https://github.com/Azure/secrets-store-csi-driver-provider-azure/pull/1057 for that change. Checking with Anish if its a backcompat change.

vishiy: confirmed that this change is back compat (meaning Id and ID would work).

vishiy (approved):

Issue #384, opened 2023-02-14
Need a way to specify the priorityClassName for the deployment


OriYosefiMSFT:
We request a lot of CPU and memory to our prom-collector deployment. When our AKS cluster is at max node capacity, the Pod would sometimes be in pending because there isn't a single node with enough resource to accommodate it. 
Setting a higher priority to the deployment would allow prom-collector to evict other nodes to make sure it can find a node with enough resources

PR #385, opened 2023-02-14
Add support for specifying priority class for the deployment

OriYosefiMSFT:
Allow specifying Pod priority
See issue #384 for details.


otelcollector/deploy/chart/prometheus-collector/templates/prometheus-collector-deployment.yaml
```diff
@@ -213,6 +213,9 @@ spec:
                 - NET_ADMIN
                 - NET_RAW
         {{- end }}
+      {{- if .Values.deploymentPriorityClass }}
```
vishiy, line 216: Ori, can you add this to our daemonset as well ?

OriYosefiMSFT: Added

PR #395, opened 2023-02-22
Update node disk usage alert to 85% to account default garbage col

Sohamdg081992:
diskusage
This pr updates the recommended alerts disk usage at node level to 85% because kubelet uses garbage collection at a high threshold of 85% by default and lowers it to the low threshold of 80%. 

Ref: https://k8s-docs.netlify.app/en/docs/concepts/cluster-administration/kubelet-garbage-collection/#:~:text=Users%20can%20adjust%20the%20following%20thresholds%20to%20tune,which%20triggers%20image%20garbage%20collection.%20Default%20is%2085%25.

### DIFF
--- a/mixins/kubernetes/rules/recording_and_alerting_rules/templates/ci_recommended_alerts.json
+++ b/mixins/kubernetes/rules/recording_and_alerting_rules/templates/ci_recommended_alerts.json
@@ -237,6 +237,25 @@
                         ]
                     },
                     {
+                        "alert": "Disk space usage for a node is greater than 85%.",
+                        "expression": "1 - avg by (namespace,cluster,job)(node_filesystem_avail_bytes{job=\"node\"}) / avg by (namespace,cluster,job)(node_filesystem_size_bytes{job=\"node\"}) > .85",
+                        "for": "PT5M",
+                        "annotations": {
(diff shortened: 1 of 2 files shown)


vishiy (approved):

mixins/kubernetes/rules/recording_and_alerting_rules/templates/ci_recommended_alerts.json
```diff
@@ -236,6 +236,25 @@
                             }
                         ]
                     },
+                    {
+                        "alert": "Disk space usage for a node is greater than 85%.",
+                        "expression": "1 - avg by (namespace,cluster,job)(node_filesystem_avail_bytes{job=\"node\"}) / avg by (namespace,cluster,job)(node_filesystem_size_bytes{job=\"node\"}) > .85",
+                        "for": "PT5M",
```
vishiy, line 242: there is no device & node label level group. So how does the customer know what disk in which node fired this alert ?

Sohamdg081992: Thanks Vishwa! I am going to correct this in the other pr.

PR #419, opened 2023-03-09
fix: enable windowsexporter and windowskubeproxy by default

bragi92:


otelcollector/configmapparser/tomlparser-default-scrape-settings.rb
```diff
@@ -19,8 +19,8 @@
 @nodeexporterEnabled = true
 @prometheusCollectorHealthEnabled = true
 @podannotationEnabled = false
-@windowsexporterEnabled = false
-@windowskubeproxyEnabled = false
```
vishiy, line 23: wont this generate error in windows ds where there is no windows exporter installed by default?

bragi92: Yeah, I was thinking to reduce a step for the customers where they only have to apply the windows exporter yaml but you're right.

I'll close this PR and update the instructions to apply the ama-metrics-settings configmap with the windows properties turned on.

PR #426, opened 2023-03-15
fix: Update role assignment to use azure monitor worksapces subscription

bragi92:
This is to fix a bug in our templates. The role assingment for the Monitoring Data Reader role needs to be created in the subscription of the Azure Monitor Workspace instead of the clusters subscription as the two resources can be in different subscriptions.


AddonArmTemplate/FullAzureMonitorMetricsProfile.json
```diff
@@ -358,7 +359,7 @@
       "name": "[parameters('roleNameGuid')]",
       "scope": "[parameters('azureMonitorWorkspaceResourceId')]",
       "properties": {
-          "roleDefinitionId": "[concat('/subscriptions/', variables('clusterSubscriptionId'), '/providers/Microsoft.Authorization/roleDefinitions/', 'b0d8363b-8ddd-447d-831f-62ca05bff136')]",
+          "roleDefinitionId": "[concat('/subscriptions/', variables('azureMonitorWorkspaceSubscriptionId'), '/providers/Microsoft.Authorization/roleDefinitions/', 'b0d8363b-8ddd-447d-831f-62ca05bff136')]",
```
vishiy, line 362: hi Kaveesh, whats this change for ?

bragi92: This is to fix a bug in our templates. The role assingment for the Monitoring Data Reader role needs to be created in the subscription of the Azure Monitor Workspace instead of the clusters subscription as the two resources can be in different subscriptions.

vishiy: ok got it.

vishiy (approved):

PR #441, opened 2023-04-01
Fix typo in configMap breaking pod annotation based scraping

zioproto:
This never worked, the typo was merged with the feature in PR #414


zioproto:
This is urgent because this file is linked to https://aka.ms/azureprometheus-addon-settings-configmap
in the docs page https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/prometheus-metrics-scrape-configuration

gracewehner (approved):

gracewehner:
Thanks @zioproto for your PR. I see now the typo is in the code also, so we will need to keep the configmap the same for now. In the next release, we will fix the code and the configmap and also make it backwards compatible with the current spelling.

PR #470, opened 2023-04-21
fix: update windows recording rules to remove duplicate entries for non running containers

bragi92:
For e.g. The following query returned a many to many matching error: 
image

because there where multiple containers within the pod stuck in initizaling state and thus did not have container_id's set:
image

Hence our recording rule for them would fail. This fix eliminates those containers which are not running at all as there is no point in getting the cpu/memory usage for them.

Link to OSS community PR fix : https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/840

### DIFF
--- a/mixins/kubernetes/rules/recording_and_alerting_rules/templates/k8s-resource-windows-cluster-rules-group-2.json
+++ b/mixins/kubernetes/rules/recording_and_alerting_rules/templates/k8s-resource-windows-cluster-rules-group-2.json
@@ -17,7 +17,7 @@
                 "rules": [
                     {
                         "record": "windows_pod_container_available_2",
-                        "expression": "windows_container_available{job = \"windows-exporter\"} * on(container_id) group_left(container, pod, namespace, cluster) max(kube_pod_container_info{job = \"kube-state-metrics\"}) by(container, container_id, pod, namespace, cluster)"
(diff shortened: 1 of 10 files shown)


gracewehner (approved):

mixins/kubernetes/rules/recording_and_alerting_rules/templates/k8s-resource-windows-cluster-rules-group-2.json
```diff
@@ -17,7 +17,7 @@
                 "rules": [
                     {
                         "record": "windows_pod_container_available_2",
-                        "expression": "windows_container_available{job = \"windows-exporter\"} * on(container_id) group_left(container, pod, namespace, cluster) max(kube_pod_container_info{job = \"kube-state-metrics\"}) by(container, container_id, pod, namespace, cluster)"
```
vishiy, line 20: @bragi92 - are these files (under mixins/kubernetes/rules/recording_and_alerting_rules/templates) used anywhere ?

bragi92: You're right. They're not used anymore. I had created them for the private preview of the windows recording rules and dashboards. I will create a PR to delete them.

vishiy: ok thanks Kaveesh.

PR #485, opened 2023-05-03
fix(managed prometheus): update azure monitor metrics windows pods cpu usage to 500m and update requests and limtis for cpu and memory

bragi92:


otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-daemonset.yaml
```diff
@@ -299,11 +299,11 @@ spec:
...
           resources:
             limits:
-              cpu: 200m
+              cpu: 500m
               memory: 1Gi
             requests:
-              cpu: 75m
-              memory: 150Mi
+              cpu: 500m
+              memory: 1Gi
```
vishiy, line 306: kaveesh - previously request wasn’t equal to limits and that’s another change as part of this pr as well?

bragi92: Yeah, I noticed that we didn't have requests == limits over here and its like this in the AKS-RP too. I'm going to be updating both (prom collector and AKS-RP) to the same values with this change

vishiy (approved):

PR #492, opened 2023-05-08
Reduce image tag length to docker limit of 128

bragi92:
Windows image tags take an extra 9 characters because of the '-ltsc2019' or '-ltsc2022' at the end of the image name. So reducing the image tag length by 9 characters to 119 so that the docker build command can run successfullly.


vishiy (approved):

gracewehner (approved):

vishiy:
thanks @bragi92

PR #493, opened 2023-05-08
Add buildver.txt file for Arc ev2 release

gracewehner:


.pipelines/deployment/arc-extension-release/ServiceGroupRoot/buildver.txt
```diff
@@ -0,0 +1 @@
+0.0.1
```
vishiy, line 1: Hi Grace, can you tell us where we use this ver file and when we increment it ?

gracewehner: Thanks Vishwa, here is the ev2 docs on it: https://ev2docs.azure.net/features/authoring/version/version.html. In our other ev2 files and for ContainerInsights, we have actually never used it since we have the image tag build id. However, I can change it so that we are putting the same as what we have in our VERSION file here?

vishiy: thanks .

vishiy (approved):

PR #495, opened 2023-05-09
Add Role assignments for new Grafana instances; Allow different RGs for Grafana and AMW

rijais:
- Grafana clusters (if created via existing bicep templates), do not allow users to access the Grafana instance. This PR adds new role assignments using user's object ID as a parameter.
- Allow Grafana to access the AMW instance if they are in different resource groups.
- Enable system assigned identity for Grafana instances created via the template to allow role assignments.


rijais:
Although I have tried out the changes - requesting to try out the set up once before approving.

bragi92 (approved):

PR #507, opened 2023-05-24
Terraform template fixes for addon

gracewehner:
Based on feedback from CSS and testing with each component in a different region

### DIFF
--- a/AddonTerraformTemplate/main.tf
+++ b/AddonTerraformTemplate/main.tf
@@ -38,14 +38,14 @@ resource "azurerm_kubernetes_cluster" "k8s" {
 resource "azurerm_monitor_data_collection_endpoint" "dce" {
   name                          = "MSProm-${var.monitor_workspace_location}-${var.cluster_name}"
   resource_group_name           = azurerm_resource_group.rg.name
-  location                      = azurerm_resource_group.rg.location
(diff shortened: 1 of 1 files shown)


AddonTerraformTemplate/main.tf
```diff
@@ -38,14 +38,14 @@ resource "azurerm_kubernetes_cluster" "k8s" {
 resource "azurerm_monitor_data_collection_endpoint" "dce" {
   name                          = "MSProm-${var.monitor_workspace_location}-${var.cluster_name}"
   resource_group_name           = azurerm_resource_group.rg.name
-  location                      = azurerm_resource_group.rg.location
+  location                      = var.monitor_workspace_location
```
vishiy, line 41: thanks for the fixes grace. So is the issue - template using RG region, instead of MAC region , there by creating no data issues for clusters where RG and MAC are in 2 different regions ?

bragi92 (approved):

AddonTerraformTemplate/main.tf
gracewehner: Yes the template fails during deployment right now if the resource group and AMW region are different

vishiy: ok thanks Grace.

PR #528, opened 2023-06-28
Arc extension: do not mount ubuntu ca-certs if k8s distro is AKS Edge Essentials

gracewehner:
AKS Edge Essentials clusters always use Mariner nodes, but have read-only host filesystems, so mounting an empty ubuntu ca-cert directory fails. Arc will provide the k8s distro as a chart value, so we will only mount the Mariner ca-certs if the distro is AKS EE.


bragi92 (approved):

rashmichandrashekar:
Thanks @gracewehner. How do these changes end up in ARC? Is it by extension release? Also, does this need any change in RP?

gracewehner:
Thanks @rashmichandrashekar, for ARC yes it'll go in with the next extension release which releases this helm chart. For RP with the addon, we don't need any changes, since regular AKS doesn't have the read-only host file systems

Sohamdg081992 (approved):

rashmichandrashekar (approved):

PR #557, opened 2023-08-09
fix: update region list for AMW location in ARM, Bicep and Policy

bragi92:
Result from RP:
image


Sohamdg081992 (dismissed):

vishiy (approved):

AddonArmTemplate/FullAzureMonitorMetricsProfile.json
```diff
@@ -10,14 +10,24 @@
       "defaultValue": "",
       "allowedValues": [
         "eastus2euap",
-        "centraluseuap",
-        "centralus",
+        "northcentralus",
```
vishiy, line 13: Kaveesh - should we also be adding fairfax region (virginia) here, as same templates will be used there too..

bragi92: Makes sense, I'll update the regions to include virginia.

bragi92: updated

bragi92: @vishiy Can you please approve the PR again when you get a chance. It should be good to merge in now.

vishiy (approved):

PR #580, opened 2023-08-31
example service monitor CR and update reference app with label

bragi92:


rashmichandrashekar:
thanks for adding this Kaveesh.

PR #590, opened 2023-09-08
Add toggle for ARC through values.yaml

gracewehner:


otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-deployment.yaml
```diff
@@ -62,7 +62,11 @@ spec:
             - name: AZMON_COLLECT_ENV
               value: "false"
             - name: AZMON_OPERATOR_ENABLED
+            {{- if or (eq .Values.AzureMonitorMetrics.ArcExtension false) (eq .Values.AzureMonitorMetrics.ArcEnableOperator true) }}
```
rashmichandrashekar, line 65: Is the ArcExtension variable always set to either true or false? Can this be nil?

gracewehner: This is a good point, thanks Rashmi. Updated this to work in case it is nil

PR #599, opened 2023-09-19
Rename target allocator

gracewehner:


gracewehner:
> Hi @gracewehner - Looks like the proxy settings are not updated with the new name, do you want to update that as well?

Thanks Rashmi, good catch, added this

PR #600, opened 2023-09-20
Add cluster scope to alert rule groups for linking them with UX

Sohamdg081992:
This adds cluster scope to alert rule groups for linking them with UX according to the latest update from AKS.


AddonArmTemplate/FullAzureMonitorMetricsProfile.json
```diff
@@ -221,7 +221,8 @@
       "properties": {
         "description": "[concat(variables('nodeRecordingRuleGroupDescription'), variables('version'))]",
         "scopes": [
-          "[parameters('azureMonitorWorkspaceResourceId')]"
+          "[parameters('azureMonitorWorkspaceResourceId')]",
```
bragi92, line 224: What benefit does adding the cluster resource id to the scope give us? Does that mean we do not need to specify the clusterName parameter?

AddonArmTemplate/FullAzureMonitorMetricsProfile.json
```diff
@@ -281,7 +282,8 @@
       "properties": {
         "description": "[concat(variables('kubernetesRecordingRuleGroupDescription'), variables('version'))]",
         "scopes": [
-          "[parameters('azureMonitorWorkspaceResourceId')]"
+          "[parameters('azureMonitorWorkspaceResourceId')]",
```
bragi92, line 285: Can you include teraform to this?

AddonArmTemplate/FullAzureMonitorMetricsProfile.json
```diff
@@ -462,7 +465,8 @@
       "properties": {
         "description": "[concat(variables('RecordingRuleGroupDescriptionWin'), variables('version'))]",
         "scopes": [
-          "[parameters('azureMonitorWorkspaceResourceId')]"
```
bragi92, line 465: Will you also be making a change to the Alerts RP from where the CLI is downloading the recording rule groups and also add this to the CLI recording rule scope -> [here](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/azuremonitormetrics/recordingrules/create.py#L31)?

Also we would need to make a similar change in the UX (Greenfield + Brownfield), can you sync with them too?

AddonArmTemplate/FullAzureMonitorMetricsProfile.json
bragi92: Based on the PR description, would these alerts show up in the Alerts Rules page on the AKS cluster if we set this scope?

Sohamdg081992: Thanks Kaveesh! Yes it would show under the cluster in the rules UI which the rules team is about to launch.

AddonArmTemplate/FullAzureMonitorMetricsProfile.json
Sohamdg081992: Thanks! My bad, I missed it. I will add it.

AddonArmTemplate/FullAzureMonitorMetricsProfile.json
bragi92: @Sohamdg081992 I am making changes in the CLI for a bug fix. I can add this scope update in that PR.

Sohamdg081992: Sounds good. Thanks Kaveesh!

AddonArmTemplate/FullAzureMonitorMetricsProfile.json
Sohamdg081992: Added terraform

bragi92 (approved):

PR #628, opened 2023-10-09
Telegraf telemetry feedback

gracewehner:


otelcollector/telegraf/telegraf-prometheus-collector-ds.conf
```diff
@@ -0,0 +1,229 @@
...
+   pid_tag = true
+   name_override = "otelcollector"
+   fieldpass = ["cpu_usage", "memory_rss"]
+   [inputs.procstat.tags]
+#    Computer = "$NODE_NAME"
+#    NodeIp = "$NODE_IP"
+    cpulimit = "$CONTAINER_CPU_LIMIT"
+    memlimit = "$CONTAINER_MEMORY_LIMIT"
+    
+    defaultscrapekubelet = "$AZMON_PROMETHEUS_KUBELET_SCRAPING_ENABLED"
```
rashmichandrashekar, line 188: Are all these setting relevant to ds?

gracewehner: No good point, we can also remove these, we only need them once from rs

rashmichandrashekar (approved):

gracewehner:
Thanks Rashmi

otelcollector/telegraf/telegraf-prometheus-collector.conf
```diff
@@ -118,6 +118,23 @@
...
+# Transforms tag and field values as well as measurement, tag and field names with regex pattern
+[[processors.regex]]
+  namepass = ["target_allocator"]
+
+  # Tag and field conversions defined in a separate sub-tables
+  [[processors.regex.tags]]
+    ## Tag to change, "*" will change every tag
+    key = "job_name"
+    ## Regular expression to match on a tag value
+    pattern = '(^(podMonitor|serviceMonitor)\/.*)|(.*)'
```
rashmichandrashekar, line 130: nit: could you pls also add a comment with what the generated podmonitor and servicemonitor names look like?

gracewehner: Yes sounds good, I'll commit this to the operator-targetallocator branch

otelcollector/telegraf/telegraf-prometheus-collector-ds.conf
```diff
@@ -0,0 +1,206 @@
...
+# Transforms tag and field values as well as measurement, tag and field names with regex pattern
+[[processors.regex]]
+  namepass = ["opentelemetry_allocator_targets"]
+
+  # Tag and field conversions defined in a separate sub-tables
+  [[processors.regex.tags]]
+    ## Tag to change, "*" will change every tag
+    key = "job_name"
+    ## Regular expression to match on a tag value
+    pattern = '(^(podmonitor|servicemonitor)\/.*)|(.*)'
```
rashmichandrashekar, line 123: @gracewehner - I was double checking the regex and looks like this seems to match any job name that has podmonitor in it? ex - anynamepodmonitor. Should we be more specific about the job names to match what TA generates?

gracewehner: Thanks Rashmi, yes so we want all job names to match so that they are replaced/overwritten and we aren't collecting the name. So we're only putting podmonitor or servicemonitor if it matches group 2, which has "podmonitor/*" will be replaced as podmonitorJob since it matches that it starts with "podmonitor/". But anynamepodmonitor doesn't match group 2 but just the right-hand side of the | (.*), so it's replaced as just 'Job' to signify it's a configmap name. So ${2} will be empty. Telegraf just won't allow empty string replacements, so I've added it to always suffix with Job

rashmichandrashekar: sounds good, thanks Grace

PR #634, opened 2023-10-18
update telemetry jitter from telegraf

vishiy:
* update telegraf jitter
* add missed exemplar setting


otelcollector/telegraf/telegraf-prometheus-collector.conf
```diff
@@ -27,6 +27,7 @@
   mode = "$MODE"
   winmode = "$WINMODE"
   macmode = "$MAC"
+  opt = "true"
```
rashmichandrashekar, line 30: Thanks for fixing this Vishwa! Is this dim just to be able to query for clusters that have this new conf rolled out?

rashmichandrashekar (approved):

PR #639, opened 2023-10-18
Adding ama-metrics to customer control plane

sgeannina:
Pending changes:
* Change address "replace" to use regex
* Remove IP address from instance
* Show Pod name
* Filter the CPU and memory metrics for the components.  Even if the cpu/memory metrics are present in the keep list, we don't want to expose these to customers.
* Currently all targets are disabled (easier for testing). By default we need apiserver and etcd enabled.

New:
  -- Remove scraping interval from settings and keep it to 30s
  -- Remove prometheus collector health from ccp
  -- Add prefix to all metrics: control-plane


otelcollector/build/linux/Dockerfile-mac
```diff
@@ -0,0 +1,182 @@
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/oss/go/microsoft/golang:1.19 as fluent-bit-builder
```
sgeannina, line 1: I'll remove this file in the final PR

otelcollector/configmapparser/default-prom-configs/clusterautoscalerDefault.yml
```diff
@@ -0,0 +1,28 @@
+  scrape_configs:
+  - job_name: cluster-autoscaler
+    scrape_interval: $$SCRAPE_INTERVAL$$
+    follow_redirects: false
```
sgeannina, line 4: would these files be replaced by the new configmap?

otelcollector/configmapparser/default-prom-configs/clusterautoscalerDefault.yml
```diff
@@ -0,0 +1,28 @@
...
+      - source_labels: [ __meta_kubernetes_pod_annotation_aks_prometheus_io_scrape ]
+        action: keep
+        regex: true
+      - source_labels: [ __meta_kubernetes_pod_annotation_aks_prometheus_io_path ]
+        action: replace
+        target_label: __metrics_path__
+        regex: (.+)
+      - source_labels: [ __address__, __meta_kubernetes_pod_annotation_aks_prometheus_io_port ]
+        action: replace
+        separator: ":"
```
sgeannina, line 23: we need to replace the address with "regex: (.+?)(\:\d+)?" but keeping it like this while we test in standalone

otelcollector/configmapparser/default-prom-configs/clusterautoscalerDefault.yml
bragi92: Yes, I'm currently working on copying these over into my branch and creating an image based on that. I will share a PR link when its ready.

PR #654, opened 2023-10-20
Add alert for new agent version in prometheus collector

Sohamdg081992:
In promql , I found this way to find the distinct count of image version labels in  kube_pod_container_info, with a specific regex. Below is the screenshot of count of distinct regex by image. Used another count to get the count of image version.
Images

Below is screenshot of query output with single count.

singlecount


internal/alerts/example-alert-template.json
```diff
@@ -251,6 +251,24 @@
                                 "actionGroupId": "/subscriptions/13d371f9-5a39-46d5-8e1b-60158c49db84/resourceGroups/ContainerInsightsPrometheusCollector-Prod/providers/microsoft.insights/actiongroups/AMA-PrometheusCollectorAlertGroup"
                             }
                         ]
+                    },
+                    {
+                        "alert": "New agent version found for prometheus collector",
+                        "expression": "count(count (kube_pod_container_info{image=~\"mcr.microsoft.com/azuremonitor/containerinsights/ciprod/prometheus-collector.*\"})  by (image)) > 2",
```
vishiy, line 257: hi soham - why count of count? just count would do, isn't it ?

Sohamdg081992: Thanks Vishwa! I have attached the screenshot of alert query with single count. Since we need only the distinct count of prom collector agent version, we have to use another count on top of the above query. The single count query shows the number of pods with each agent version, but we don't need that count.

vishiy (approved):

PR #688, opened 2023-12-06
Drop etcd2 metrics

sgeannina:


otelcollector/configmapparser/default-prom-configs/controlplane_etcd.yml
```diff
@@ -31,7 +31,10 @@ scrape_configs:
       regex: (.*)
       target_label: instance
       action: replace
+    - source_labels: [ __meta_kubernetes_pod_name ]
```
davidkydd, line 34: just to confirm, the drop was previously being applied to both etcd1 and etcd2, now it is only being applied to etcd2?

sgeannina: there was no drop before, it was showing both, now it drops etcd2 instances

davidkydd: oh apologies got it - the newline removal at the bottom threw me, I thought these were matchers for the relabel config section below, but they are independent

davidkydd (approved):

PR #693, opened 2023-12-11
Remove apiserver-token secret

sgeannina:


.trivyignore
```diff
@@ -7,6 +7,7 @@
 # =========== HIGH ================
 # HIGH - telegraf
 GHSA-fr2g-9hjm-wr23
+CVE-2023-47108
```
davidkydd, line 10: is this intended?

davidkydd (approved):

.trivyignore
sgeannina: yes we need it for trivy to pass, this fix will be applied around Feb-March next year, I'll add a comment with more info

PR #714, opened 2024-01-09
Fix build issues

rashmichandrashekar:


.pipelines/azure-pipeline-build.yml
```diff
@@ -820,6 +820,47 @@ jobs:
         pathToPublish: '$(Build.ArtifactStagingDirectory)'
         artifactName: drop
 
+- job: Helm_Chart
```
bragi92, line 823: Is this for the deployment to the Dev Cluster?

bragi92 (approved):

vishiy (approved):

.pipelines/azure-pipeline-build.yml
rashmichandrashekar: yes, since we removed package helm chart, this was failing. added this back

PR #727, opened 2024-01-24
fix: update windows liveness timeoutSeconds, periodSeconds to 60 and reduce tasklist usage to once

bragi92:
The current liveness probe is taking more than 15 seconds to run resulting in timeouts on a lot of windows clusters. This PR aims to increase the timeout to a more reasonable value. In the future when we move to golang for liveness probe in windows we can reduce the timeout seconds duration back to 15.

Sample liveness probe failure with the new update:

image

### DIFF
--- a/otelcollector/build/windows/scripts/livenessprobe.cmd
+++ b/otelcollector/build/windows/scripts/livenessprobe.cmd
@@ -18,15 +18,15 @@ for /f "tokens=*" %%a in ('tasklist /fo "table"') do (
     set "output=%%a"
 
     REM Check for MetricsExtension.Native.exe
-    echo !output! | findstr /i "MetricsExtension.Native.exe" > nul
+    echo !output! | findstr /i "MetricsExtension" > nul
     if !errorlevel! equ 0 set MetricsExtension=true
 
     REM Check for MonAgentLauncher.exe
-    echo !output! | findstr /i "MonAgentLauncher.exe" > nul
+    echo !output! | findstr /i "MonAgentLauncher" > nul
     if !errorlevel! equ 0 set MonAgentLauncher=true
 
     REM Check for otelcollector.exe
-    echo !output! | findstr /i "otelcollector.exe" > nul
+    echo !output! | findstr /i "otelcollector" > nul
     if !errorlevel! equ 0 set otelcollector=true
 )
 
--- a/otelcollector/build/windows/scripts/livenessprobe.cmd
+++ b/otelcollector/build/windows/scripts/livenessprobe.cmd
@@ -1,18 +1,37 @@
 rem Get the current date and time
 setlocal enableextensions
 setlocal enabledelayedexpansion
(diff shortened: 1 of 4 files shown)


otelcollector/build/windows/scripts/livenessprobe.cmd
```diff
@@ -1,18 +1,38 @@
+@echo off
 rem Get the current date and time
 setlocal enableextensions
 setlocal enabledelayedexpansion
-for /f %%x in ('wmic path win32_utctime get /format:list ^| findstr "="') do (
-    set %%x)
+for /f %%x in ('wmic path win32_utctime get /format:list ^| findstr "="') do ( set %%x )
```
vishiy, line 5: do we use %%x anywhere after we set ?

otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-daemonset.yaml
```diff
@@ -396,9 +396,9 @@ spec:
                 - cmd
                 - /c
                 - C:\opt\microsoft\scripts\livenessprobe.cmd
-            periodSeconds: 15
+            periodSeconds: 60
```
vishiy, line 399: kaveesh - does this interfere/stamp with any waits that we are doing in startup script for initial startup ?

bragi92: I don't think this will interfere with it. The liveness probe is set to start after a delay of 5 mins right now. Setting period seconds to 60 seconds will make it so that the probe is called every 60 seconds instead of the 15 second interval it was being called at originally. Since I was increasing the timeout duration to 60 seconds I think it makes sense to call the next liveness probe at least 60 seconds later, giving enough time for the previous one to finish/timeout.

otelcollector/build/windows/scripts/livenessprobe.cmd
bragi92: For each line in the output of the wmic command that contains an equal sign, the loop variable %%x is set to the content of that line, effectively setting environment variables based on the output of the wmic command.

For example, if the output of wmic path win32_utctime get /format:list includes a line like Year=2024, then %%x would be set to Year=2024, and the line set %%x would set the environment variable %Year% to 2024 and those variables are being used in the code below to calculate the epochTimeNow.

otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-daemonset.yaml
vishiy: thanks

otelcollector/build/windows/scripts/livenessprobe.cmd
vishiy: thanks

vishiy (approved):

PR #729, opened 2024-01-24
[infra] Add remote ACR registry cache to improve linux build times

gracewehner:
- `--cache-to type=registry,ref=<ACR>,mode=max` pushes an image that is a cache of each image layer. The cache is separate from the built image
- `--cache-from type=registry,ref=<ACR>` will pull any image layers that haven't been changed and don't need to be re-built
- I have a different cache image for each of the 3 linux images. With this setup, the cache image will be overwritten after each image build at the location public/azuremonitor/containerinsights/cidev/prometheus-collector/images/cache:<prometheuscollector, targetallocator, or cfg>. This does not have an MCR sync rule, but it also doesn't have an ACR delete rule set up either since it will be overwritten frequently.
- We could also have the cache image be pushed only for merges to main, but pulled for all builds (PRs and branches)

### DIFF
--- a/.pipelines/azure-pipeline-build.yml
+++ b/.pipelines/azure-pipeline-build.yml
@@ -337,7 +337,6 @@ jobs:
 
     - bash: |
         mkdir -p $(Build.ArtifactStagingDirectory)/linux
-        docker version
 
         # Necessary due to necessary due to https://stackoverflow.com/questions/60080264/docker-cannot-build-multi-platform-images-with-docker-buildx
         sudo apt-get update && sudo apt-get -y install qemu binfmt-support qemu-user-static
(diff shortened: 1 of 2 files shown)


Sohamdg081992 (approved):

.pipelines/azure-pipeline-build.yml
```diff
@@ -345,7 +345,7 @@ jobs:
 
         docker buildx create --name dockerbuilder
         docker buildx use dockerbuilder
-        docker buildx build . --platform=linux/amd64,linux/arm64 --file ./build/linux/Dockerfile -t $(LINUX_FULL_IMAGE_NAME) --metadata-file $(Build.ArtifactStagingDirectory)/linux/metadata.json --push
+        docker buildx build . --platform=linux/amd64,linux/arm64 --file ./build/linux/Dockerfile -t $(LINUX_FULL_IMAGE_NAME) --metadata-file $(Build.ArtifactStagingDirectory)/linux/metadata.json --push --cache-to type=registry,ref=$(ACR_REGISTRY)$(ACR_REPOSITORY)/cache:prometheuscollector,mode=max --cache-from type=registry,ref=$(ACR_REGISTRY)$(ACR_REPOSITORY)/cache:prometheuscollector
```
Sohamdg081992, line 348: As you have mentioned in PR, having the cache image pushed only for merges to main but pulled for all builds, would also be a good idea since only the necessary artifacts will be cached and pulling from cache will be consistent across the builds.

vishiy, line 348: Hi Grace - does this help much ? Because every build creates a  different manifest (due to changes in our layer), which cannot be reused by future builds. Isn't it ? Also, when these build tasks are done in parallel, they cannot be used across each, isn't it ?

I see there is no MCR reconciliation for cache, bit will we be flagged by s360 for stale manifests in our ACR (the reason due to which we currently clean up in 2 hours...)

vishiy: Thanks Grace.

PR #733, opened 2024-01-26
[Arc] Upgrade node-exporter chart from 4.21.0 to 4.26.0

gracewehner:
- The newest node-exporter chart has a service.enabled field added in values.yaml. We can utilize this to allow customers to change this to false. This option was added for those who don't use serviceMonitor service discovery to scrape node exporter: https://github.com/prometheus-community/helm-charts/pull/3749
- This also updates the image from 1.6.0 to 1.7.0 which is available from upstream team:https://mcr.microsoft.com/v2/oss/prometheus/node-exporter/tags/list

### DIFF
--- a/otelcollector/deploy/dependentcharts/prometheus-node-exporter/templates/_helpers.tpl
+++ b/otelcollector/deploy/dependentcharts/prometheus-node-exporter/templates/_helpers.tpl
@@ -55,8 +55,8 @@ release: {{ .Release.Name }}
 Selector labels
 */}}
 {{- define "prometheus-node-exporter.selectorLabels" -}}
-app: {{ template "prometheus-node-exporter.name" . }}
-release: {{.Release.Name }}
+app.kubernetes.io/name: {{ include "prometheus-node-exporter.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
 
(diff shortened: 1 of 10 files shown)


otelcollector/deploy/dependentcharts/prometheus-node-exporter/templates/_helpers.tpl
```diff
@@ -55,8 +55,8 @@ release: {{ .Release.Name }}
 Selector labels
 */}}
 {{- define "prometheus-node-exporter.selectorLabels" -}}
-app: {{ template "prometheus-node-exporter.name" . }}
-release: {{.Release.Name }}
+app.kubernetes.io/name: {{ include "prometheus-node-exporter.name" . }}
```
vishiy, line 58: this will break upgrades for node exporter in ARC, as selector labels are changing (from app to app.kubernetes.io) ?

otelcollector/deploy/dependentcharts/prometheus-node-exporter/templates/_helpers.tpl
```diff
@@ -55,8 +55,8 @@ release: {{ .Release.Name }}
...
+app.kubernetes.io/instance: {{ .Release.Name }}
```
vishiy, line 59: same here - selector label change breaks upgrade...

vishiy (approved):

PR #734, opened 2024-01-26
Add telemetry for collector and addon token adaptor

Sohamdg081992:
This change adds the telemetry for collector and addon token adaptor. 
It also updates the pod name from the podRef from Cadvisor Json output since the podname environment variable points to the pod from which the cadvisor endpoint is curl-ed.

test branch: https://github-private.visualstudio.com/azure/_build/results?buildId=74436&view=results

Screenshot showing RefPodName of ama-metrics-ksm container:
PodRefName

Screenshot showing RefPodName of addon-token-adapter container:

RefPodToken

### DIFF
--- a/otelcollector/fluent-bit/src/telemetry.go
+++ b/otelcollector/fluent-bit/src/telemetry.go
@@ -406,6 +406,9 @@ func SendCoreCountToAppInsightsMetrics() {
 // Struct for getting relevant fields from JSON object obtained from cadvisor endpoint
 type CadvisorJson struct {
 	Pods []struct {
+		PodRef struct {
+			PodName string `json:"name"`
+		} `json:"podRef"`
 		Containers []Container `json:"containers"`
 	} `json:"pods"`
 }
@@ -433,6 +436,7 @@ func SendContainersCpuMemoryToAppInsightsMetrics() {
 	ksmTelemetryTicker := time.NewTicker(time.Second * time.Duration(ksmAttachedTelemetryIntervalSeconds))
 	for ; true; <-ksmTelemetryTicker.C {
 		for podId := 0; podId < len(p.Pods); podId++ {
+			podName := strings.TrimSpace(p.Pods[podId].PodRef.PodName)
 			for containerId := 0; containerId < len(p.Pods[podId].Containers); containerId++ {
 				container := p.Pods[podId].Containers[containerId]
 				containerName := strings.TrimSpace(container.Name)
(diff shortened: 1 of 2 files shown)


otelcollector/fluent-bit/src/telemetry.go
```diff
@@ -443,18 +447,22 @@ func SendContainersCpuMemoryToAppInsightsMetrics() {
...
+					GetAndSendContainerCPUandMemoryFromCadvisorJSON(container, ksmCpuMemoryTelemetryName, "MemKsmRssBytes", podName)
 				case "targetallocator":
-					GetAndSendContainerCPUandMemoryFromCadvisorJSON(container, "taCPUUsage", "taMemRssBytes")
+					GetAndSendContainerCPUandMemoryFromCadvisorJSON(container, "taCPUUsage", "taMemRssBytes", podName)
 				case "config-reader":
-					GetAndSendContainerCPUandMemoryFromCadvisorJSON(container, "cnfgRdrCPUUsage", "cnfgRdrMemRssBytes")
+					GetAndSendContainerCPUandMemoryFromCadvisorJSON(container, "cnfgRdrCPUUsage", "cnfgRdrMemRssBytes", podName)
+				case "addon-token-adapter":
+					GetAndSendContainerCPUandMemoryFromCadvisorJSON(container, "adnTknAdtrCPUUsage", "adnTknAdtrMemRssBytes", podName)
+				case "prometheus-collector":
```
gracewehner, line 457: Thanks Soham, so with the podName added, we will be able to differentiate the prometheus-collector rs container vs the prometheus-collector ds containers?

Sohamdg081992: Yes, previously it was outputting  the pod name from which the cadvisor endpoint is curl-ed. It was getting that value set in environment variable when the telemetry client is initialized.

otelcollector/fluent-bit/src/telemetry.go
```diff
@@ -463,6 +471,8 @@ func GetAndSendContainerCPUandMemoryFromCadvisorJSON(container Container, cpuMet
 
 	// Abbreviated properties to save telemetry cost
 	metricTelemetryItem.Properties[memMetricName] = fmt.Sprintf("%d", int(memoryRssBytesLinux))
+	// Updating the pod name from the podRef from Cadvisor Json output since the podname environment variable points to the pod from which the cadvisor endpoint is curl-ed
+	CommonProperties["podname"] = fmt.Sprintf("%s", podName)
```
rashmichandrashekar, line 475: Thanks for fixing this Soham. podname already exists today. I think none of our dashboards use this and today it just storing the value of the pod in which the plugin is running and sending data from(which is ama-metrics-node-* pod). But is it possible to rename it to something else so that it is not confusing? Because at every other place podname is always the pod from which the telemetry is being sent.

Sohamdg081992: Thanks Rashmi! Yes I was thinking about this to have a different dimension name instead of updating the existing podName dimension. Let me update this.

rashmichandrashekar (approved):

PR #751, opened 2024-02-20
Relabel host

sgeannina:
Mask host value with hash or alias 

Metrics without a host won't be affected
image

Metrics with a host display a hash or alias value
image


otelcollector/configmapparser/default-prom-configs/controlplane_apiserver.yml
```diff
@@ -36,6 +36,23 @@ scrape_configs:
       target_label: instance
       action: replace
   metric_relabel_configs:
+    # Generate host alias
+    - source_labels: [ host ]
+      action: hashmod
+      regex: (.+)
```
sgeannina, line 42: this filter doesn't really work as expected and it generates a hash for all metrics, the column is dropped at the end anyway so this behaviour is not visible. The replace logic has to check both [hostalias, host] as hashed alias will always have a value but host will be empty for metrics without this label.

bragi92 (approved):

PR #756, opened 2024-02-27
Update recommendedMetricAlerts.json

jacobaccumulus:
Parentheses were missing in one alert, causing deployment to fail with a missing parentheses error.

Tested using azurerm_resource_group_template_deployment terraform resource and http data block pointing at this repo (fails with missing parentheses error) and then my fork (succeeds, all rules created).

I understand this is likely a generated file so it may just point to a bug somewhere else but wanted to raise attention in order to get it fixed.

### DIFF
--- a/GeneratedMonitoringArtifacts/Default/recommendedMetricAlerts.json
+++ b/GeneratedMonitoringArtifacts/Default/recommendedMetricAlerts.json
@@ -814,7 +814,7 @@
                     },
                     {
                         "alert": "Average PV usage is greater than 80%",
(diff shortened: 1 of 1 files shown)


Sohamdg081992 (approved):

Sohamdg081992:
Thanks @jacobaccumulus for catching this. This is a copy paste bug which got introduced in the Arm template and also in bicep(since it is generated from arm). I have double checked all the recommended alerts in our repo(both arm and bicep) for syntax errors in Grafana and this was the only one. I have made a pr to update both the templates here https://github.com/Azure/prometheus-collector/pull/757. Closing this PR.

PR #771, opened 2024-03-08
[ArcA Only] Fix custom environment value on ArcA environment

XiangyuKuangMSFT:
Prom Collector on ArcA environment got wrong "customEnvironment" value since a new if statement added before "isArcACluster" check


ms-hujia (approved):
LGTM

PR #776, opened 2024-03-12
Windows tests for container status

bragi92:
- [Container Status] Add windows containers to test for all containers running
- [Container Status] Add windows containers to test that all processes are running on container
- [Container Status] Add windows containers to test that there are no errors in the logs


gracewehner (approved):
Thanks Kaveesh, LGTM

PR #778, opened 2024-03-13
build: onebranch migration

gracewehner:


vishiy (dismissed):

vishiy:
Grace - looks like there is a new CVE that you would need to include in trivy  for tgf.. hopefully when we merge tgf removal , it will get fixed by itself :)

vishiy (approved):

PR #802, opened 2024-03-26
Query tests

bragi92:
[Query] up=1 for all default job
[Query] metric existing for each default job
[Query] cluster, job, and instance labels exist on metrics

### DIFF
--- a/otelcollector/test/querymetrics/query_metrics_test.go
+++ b/otelcollector/test/querymetrics/query_metrics_test.go
@@ -1,8 +1,11 @@
 package querymetrics
 
 import (
+	"fmt"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/prometheus/common/model"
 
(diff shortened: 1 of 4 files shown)


otelcollector/test/querymetrics/query_metrics_test.go
```diff
@@ -11,11 +14,359 @@ import (
...
+						found = true
+						break
+					}
+				}
+				Expect(found).To(BeTrue(), fmt.Sprintf("Expected metric %q not found", metric))
+			}
+		})
+	})
+
+	It("should return the expected results for specified metrics in the default job 'kubelet'", func() {
```
gracewehner, line 101: Thanks Kaveesh! My only suggestion would be to turn these tests into a DescribeTable() test with the job name and expected metric list as the parameters and an Entry for each job

bragi92: Thanks Grace, updated PR to reflect this.

PR #804, opened 2024-03-26
Fix: Change logging from error to info for missing configmap settings to not have errors for older configmaps

gracewehner:
Changed to info for all defaults to keep consistency, but it was the kappie and podannotation defaults that we were seeing this with since they were added after the original configmap


vishiy (approved):
thanks Grace.

bragi92 (approved):

rashmichandrashekar (approved):

PR #809, opened 2024-04-02
Upgrade: fluent-bit from 2.0.9 to 2.1.10

gracewehner:
- Add new .so files for upgraded fluent-bit
- Fix container log collection path for Windows
- Add Windows container CPU and memory usage telemetry for prometheus-collector and addon-token-adapter-win containers
- Only send keeplistregex and scrapeinterval telemetry from the replicaset since these env vars are the same no matter if it's on the daemonset or replicaset
- Add app insights key env var in Dockerfile to non-distroless build for whenever building without the distroless image
- Fix bug in build for image check before deploying to the dev clusters

### DIFF
--- a/otelcollector/build/linux/Dockerfile
+++ b/otelcollector/build/linux/Dockerfile
@@ -82,101 +82,101 @@ RUN chmod 775 $tmpdir/*.sh;
 RUN sync;
 RUN $tmpdir/setup.sh ${TARGETARCH}
 # If wanting to run without distroless, uncomment this line and comment everything after
-#CMD [ "/opt/main.sh" ]
-
-FROM mcr.microsoft.com/cbl-mariner/distroless/base:2.0
-# Below is for ContainerInsightsPrometheusCollector-Prod AppInsights Resource
-ENV APPLICATIONINSIGHTS_AUTH_PUBLIC MWNkYTMxMTItYWY1Ni00ZmNiLWI4MDQtZjg5NDVhYTFjYjMy
-# Below is for ContainerInsightsPrometheusCollector-Fairfax AppInsights Resource
-ENV APPLICATIONINSIGHTS_AUTH_USGOVERNMENT ZmRjMTE0MmUtY2U0YS1mNTFmLWE4M2EtODBjM2ZjNDYwNGE5
-# Below is for ContainerInsightsPrometheusCollector-Mooncake AppInsights Resource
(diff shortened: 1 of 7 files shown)


Sohamdg081992 (approved):

otelcollector/build/linux/Dockerfile
```diff
@@ -152,7 +162,7 @@ COPY --from=builder /usr/lib/libre2.so.0a /usr/lib/libstdc++.so.6 /usr/lib/libgc
...
 # fluent-bit dependencies
 # libssl.so.1.1 & libcrypto.so.1.1 are already available with openssl in distroless and copying them over causes FIPS HMAC verification failures
-COPY --from=builder  /lib/libyaml-0.so.2 /lib/libsystemd.so.0 /lib/libm.so.6 /lib/libgcc_s.so.1 /lib/liblzma.so.5 /lib/liblz4.so.1 /lib/libcap.so.2 /lib/libgcrypt.so.20 /lib/libgpg-error.so.0 /lib/
+COPY --from=builder /lib/libyaml-0.so.2 /lib/libsystemd.so.0 /lib/libcurl.so.4 /lib/libm.so.6 /lib/libz.so.1 /lib/libzstd.so.1 /lib/libsasl2.so.3 /lib/libgcc_s.so.1 /lib/libc.so.6 /lib/liblzma.so.5 /lib/liblz4.so.1  /lib/libcap.so.2 /lib/libgcrypt.so.20 /lib/libnghttp2.so.14 /lib/libssh2.so.1 /lib/libgssapi_krb5.so.2 /lib/libresolv.so.2 /lib/libgpg-error.so.0 /usr/lib/libkrb5.so.3 /usr/lib/libk5crypto.so.3 /usr/lib/libcom_err.so.2 /usr/lib/libkrb5support.so.0 /lib/
```
Sohamdg081992, line 165: I see 2 more dependencies /lib64/ld-linux-x86-64.so. and /lib64/ld-linux-x86-64.so.2 while running ldd /usr/bin/fluent-bit. Do we exclude them since they are the only ones not under the lib directory?

gracewehner: Right the ones that don't have `=>` pointing to the file location when running `ldd` don't need to be copied over, as these are always in the distroless image

Sohamdg081992: Sounds good. Thanks Grace!

bragi92 (approved):

vishiy (approved):

rashmichandrashekar (approved):

PR #817, opened 2024-04-11
Add back target UX

gracewehner:
pkgtranslatorprometheus is in both the otelcollector and prometheus repos. This is needed to be custom by us until the planned fix is merged back to change 'MustRegister' to 'Register' to not cause a panic.


otelcollector/prom-config-validator-builder/main.go
```diff
@@ -25,6 +25,7 @@ type OtelConfig struct {
 		Prometheus struct {
 			Config          interface{} `yaml:"config"`
 			TargetAllocator interface{} `yaml:"target_allocator"`
+			PrometheusAPIServerExtension interface{} `yaml:"prometheus_api_server_extension"`
```
rashmichandrashekar, line 28: @gracewehner - Is this required here as well? https://github.com/Azure/prometheus-collector/blob/320444c6ec2f0298fcb4540b6e0bc32dcaedd0a7/otelcollector/configuration-reader-builder/main.go#L35C2-L40C17

PR #828, opened 2024-04-12
[fix] Minor fix in onboarding templates

Sohamdg081992:
This change is a minor fix for bicep template


ArcBicepTemplate/FullAzureMonitorMetricsProfile.bicep
```diff
@@ -1,48 +1,7 @@
...
-  'switzerlandwest',
-  'japaneast',
-  'swedencentral',
-  'canadaeast',
-  'norwaywest',
-  'southindia',
-  'australiaeast',
-  'swedensouth',
-  'usgovvirginia'
+  'eastus2euap', 'northcentralus', 'eastus', 'australiacentral', 'australiasoutheast', 'brazilsouth', 'canadacentral', 'centralindia', 'centralus', 'eastasia', 'eastus2', 'northeurope', 'norwayeast', 'southafricanorth', 'southcentralus', 'southeastasia', 'uaenorth', 'uksouth', 'westcentralus', 'westeurope', 'westus', 'westus2', 'francesouth', 'germanywestcentral', 'israelcentral', 'italynorth', 'japanwest', 'koreacentral', 'switzerlandnorth', 'brazilsoutheast', 'francecentral', 'ukwest', 'koreasouth', 'switzerlandwest', 'japaneast', 'swedencentral', 'canadaeast', 'norwaywest', 'southindia', 'australiaeast', 'swedensouth', 'usgovvirginia'
```
bragi92, line 4: Could you please modify all templates (ARM, Bicep, Policy, Terraform, etc.) to eliminate the allowed region list entirely? This change would streamline our processes by removing the need for constant updates whenever new regions are added.

Even without a predefined list, if a customer selects an unsupported region, they'll receive an error message from the Resource Provider indicating that the region isn't supported. This should help with maintenance overhead of these files.

Sohamdg081992: Thanks Kaveesh! It makes sense, let me update this.

Sohamdg081992: I have updated the templates now. Please take a look.

bragi92 (approved):

PR #842, opened 2024-04-19
Add life cycle metadata to container image

Sohamdg081992:
This change adds life cycle metadata to the container image.

Followed this document: https://eng.ms/docs/more/containers-secure-supply-chain/lifecycle-and-lineage

This adds metadata end-of-life annotation at the image build time so that it does not get flagged later in downstream.

### DIFF
--- a/.pipelines/azure-pipeline-build.yml
+++ b/.pipelines/azure-pipeline-build.yml
@@ -2,6 +2,7 @@ trigger:
   branches:
     include:
     - main
+    - AddLifeCycleImage
 
 pr:
   autoCancel: true
--- a/.pipelines/azure-pipeline-build.yml
+++ b/.pipelines/azure-pipeline-build.yml
@@ -404,6 +404,9 @@ stages:
             --artifact-type 'application/vnd.cncf.notary.signature' \
             ./payload.json:application/cose \
             -a "io.cncf.notary.x509chain.thumbprint#S256=[\"79E6A702361E1F60DAA84AEEC4CBF6F6420DE6BA\"]"
+          oras attach $(LINUX_FULL_IMAGE_NAME) \
+            --artifact-type 'application/vnd.microsoft.artifact.lifecycle' \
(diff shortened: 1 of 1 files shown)


gracewehner:
Thanks Soham, so is the end-of-life the time when they won't be flagged anymore? Should we set this to be the current time the image is built or +1h so that these images aren't flagged before 30 days?

Sohamdg081992:
> Thanks Soham, so is the end-of-life the time when they won't be flagged anymore? Should we set this to be the current time the image is built or +1h so that these images aren't flagged before 30 days?

Thanks Grace! Yes that will be good to set this at the current time it is built + 1h(buffer). Let me update this.

.pipelines/azure-pipeline-build.yml
```diff
@@ -404,6 +404,9 @@ stages:
             --artifact-type 'application/vnd.cncf.notary.signature' \
             ./payload.json:application/cose \
             -a "io.cncf.notary.x509chain.thumbprint#S256=[\"79E6A702361E1F60DAA84AEEC4CBF6F6420DE6BA\"]"
+          oras attach $(LINUX_FULL_IMAGE_NAME) \
+            --artifact-type 'application/vnd.microsoft.artifact.lifecycle' \
+            --annotation "vnd.microsoft.artifact.lifecycle.end-of-life.date=$(date -u -d '+1 hour' +"%Y-%m-%dT%H:%M:%SZ")"
```
vishiy, line 409: Any reason why we are adding 1 hour ? We can just use current time , or start of the day time (00:00:00). It also allows past time as well, so that means {currentTime -1} hour would work even better.

Sohamdg081992: Thanks @vishiy ! I was thinking adding 1 hour would account for some buffer time after the build step succeeds but I agree it will be better to do past time (1 hour) instead so that our images are not flagged. Making this change now.

Sohamdg081992: Updated the annotation. Thanks!

vishiy (approved):

PR #930, opened 2024-07-02
Test: unit tests and some small fixes for configmap processing

gracewehner:
# PR Description
- Moved all file names to be variables in the `definitions.go` file and replaced any hardcoded paths with the variables. Found this is standard practice for testing with temp files, so that for tests, we can create a temp file with a different path, but still use the same code
- Pod Annotations
   - Removed parsing an enabled setting since it's a special case where anything in the `podannotationsnamespaceregex` enables it
   - Made some fixes for the parsing of the config file and setting the env var for if it's enabled or not based on if the namespace regex is present/valid or not
   - Changed the keeplist setting for pod annotations to be a separate setting from the `podannotationsnamespaceregex` to match the current ruby implementation
- Debug Mode
   - Changed from using the global variable in the file for if it's enabled or not to having the function return if it's enabled or not, since this was messing with the tests
   - The helper function for reading the env vars from a file and writing to .bashrc do not expect `export` in the file, so changed this to match the other files
 - Scrape Interval and Regex Keep List (still need to be fully reviewed, but had found these bugs from testing pod annotations)
   - Scrape Interval: Fixed if statements for if configmap doesn't exist, default 30s weren't being set
   - KeepList: Using `+=` was causing issues with ordering on where the keep list regex was being added
   - If the configmap didn't exist, the minimal ingestion profile settings weren't getting added
- Ginkgo unit tests
   - Unit tests for debug mode and pod annotation individual files to check if env vars are getting set correctly based on the configmap settings given
   - `Configmapparser_test.go`:
       - Example tests and outline of other tests needed
          - Sets up the env variables for the scenario, the temp files for the settings configmap with the settings to test, the temp files for where the env vars will be written to for .bashrc, and the file that has the expected Prometheus config that will be generated
          - Runs the `Configmapparser()` to test the processing
          - Checks all env vars are set to what is expected
          - Checks the regex keeplist and scrape interval hashmaps have the expected values
          - Checks that the generated Prometheus config file matches the expected file
       - Scenarios needed to test: Linux replicaset, linux daemonset, and windows daemonset for
          - No configmap exists
          - Configmap exists but doesn't have the setting values
          - Configmap exists and have changed the values to be non-default


bragi92 (approved):

Sohamdg081992 (approved):

otelcollector/shared/configmap/mp/tomlparser-default-scrape-settings.go
```diff
@@ -40,7 +39,6 @@ func (fcl *FilesystemConfigLoader) ParseConfigMapForDefaultScrapeSettings() (map
 	config["kubestate"] = "true"
 	config["nodeexporter"] = "true"
 	config["prometheuscollectorhealth"] = "false"
-	config["podannotation"] = "false"
```
Sohamdg081992, line 43: just a curious question, why are  we removing podannotation setting from here?

PR #962, opened 2024-08-19
Config Reader sidecar memory leak fix

rashmichandrashekar:
# PR Description

# New Feature Checklist

[comment]: # (The below checklist is for code changes. Not all boxes necessarily need to be checked. Build, doc, and template changes do not need to fill out the checklist.)
# Tests Checklist

- [ ] Have end-to-end Ginkgo tests been run on your cluster and passed? To bootstrap your cluster to run the tests, follow [these instructions](/otelcollector/test/README.md#bootstrap-a-dev-cluster-to-run-ginkgo-tests).
    - [ ] `operator`
    - [ ] `windows`
    - [ ] `arm64`
    - [ ] `arc-extension`
    - [ ] `fips`
- [ ] Have new tests been added? For features, have tests been added for this feature? For fixes, is there a test that could have caught this issue and could validate that the fix works?
  - [ ] Is a new scrape job needed?
    - [ ] The scrape job was added to the folder [test-cluster-yamls](/otelcollector/test/test-cluster-yamls/) in the correct configmap or as a CR. 
  - [ ] Was a new test label added?
    - [ ] A string constant for the label was added to [constants.go](/otelcollector/test/utils/constants.go).
    - [ ] The label and description was added to the [test README](/otelcollector/test/README.md).
    - [ ] The label was added to this [PR checklist](/.github/pull_request_template).
    - [ ] The label was added as needed to [testkube-test-crs.yaml](/otelcollector/test/testkube/testkube-test-crs.yaml).
  - [ ] Are additional API server permissions needed for the new tests?
    - [ ] These permissions have been added to [api-server-permissions.yaml](/otelcollector/test/testkube/api-server-permissions.yaml).
  - [ ] Was a new test suite (a new folder under `/tests`) added?
    - [ ] The new test suite is included in [testkube-test-crs.yaml](/otelcollector/test/testkube/testkube-test-crs.yaml).


bragi92 (approved):

otelcollector/configuration-reader-builder/main.go
```diff
@@ -209,6 +209,7 @@ func taHealthHandler(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(status)
 		fmt.Fprintln(w, message)
 	}
+	defer resp.Body.Close()
```
bragi92, line 212: @rashmichandrashekar Thanks for fixing this. Can you also add a condition like this [here](https://github.com/Azure/prometheus-collector/blob/main/otelcollector/shared/process_utilities.go#L418) ?

gracewehner, line 212: Thanks Rashmi, I've seen in the past when `http.Get` returns an error then `resp` is nil and `defer resp.Body.Close()` causes a segfault. Could we add `if resp != nil` around this?

rashmichandrashekar: Thanks @bragi92. Looking at this code I am wondering, this http request is not closed as well, which might lead to mem leaks. Did you see that behavior?

rashmichandrashekar: Thanks @gracewehner. Fixed.

bragi92: No, I haven't seen memory leaks yet for this. This has been running on our ci-prod cluster for 11 days now but surpsingly no extra memory usage ~ maybe its a very small leak? I think it should be closed either way though so that it doesn't have a chance of misbehaving.

rashmichandrashekar: Thanks @gracewehner. Fixed

rashmichandrashekar: Yes, It might be very small depending on the size of the response. I think we need to look at all the http calls we are making and make sure we are closing them. Lets sync during scrum, we can look at memory patterns too.

rashmichandrashekar: So, from the code looks like the liveness check for healthz endpoint might not cause memory leak because it is only called a few times and then we give up. So, it doesnt add up over time. But I added a close for this anyway and dont see any other http calls other than these.

otelcollector/shared/process_utilities.go
```diff
@@ -406,16 +406,18 @@ func WaitForTokenAdapter(ccpMetricsEnabled string) {
 	}
 	waitedSecsSoFar := 1
 
+	var resp *http.Response
+
 	for {
 		if waitedSecsSoFar > tokenAdapterWaitSecs {
-			if _, err := http.Get("http://localhost:9999/healthz"); err != nil {
+			if resp, err = http.Get("http://localhost:9999/healthz"); err != nil {
```
vishiy, line 413: Hi Rashmi - instead of using the global http client, can we just create a new explicit client, and when doing so, specify like a 2s timeout, so we don't piggy-back on the global default client's no timeout ?  please look for this pattern across everywhere where we are using http.<...> to make them not use the default client...

gracewehner:
Thanks Rashmi, I added the task we discussed about the linter to our board with some details [here](https://dev.azure.com/msazure/InfrastructureInsights/_sprints/taskboard/Prometheus-agent/InfrastructureInsights/Dilithium/CY24Q3/Monthly/08%20Aug%20(Aug%2001%20-%20Aug%2031)?workitem=29148875).

otelcollector/shared/process_utilities.go
rashmichandrashekar: thanks Vishwa, added. Testing them out.

bragi92 (approved):

otelcollector/shared/process_utilities.go
vishiy: thanks.

vishiy (approved):

PR #999, opened 2024-10-18
feat: allow json string as input

monteiro-renato:
Hey 👋,

It would be nice if this tool could also support a JSON string as input.
For my use-case, it's easier to Marshal the [prometheus-operator PrometheusRuleSpec](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/prometheusrule_types.go#L50-L55) and then feed the JSON string directly to this tool instead of having to convert the JSON to YAML, store it in a file and only then call the tool.


monteiro-renato:
Still interested but awaiting feedback from codeowners.

vishiy:
@moshemal - can you look at this PR ?

vishiy:
@moshemal

monteiro-renato:
Hey @moshemal  👋,
Thanks for looking into this.

I guess it might work since YAML is a superset of JSON but that's not exactly my main problem. The tool currently tries to read the input from a file and that's what I would like to avoid. I would like to "pipe" a JSON string directly into the tool instead.

For a bit more context, we depend on a tool (a k8s controller) that generates [PrometheusRule](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PrometheusRule) Custom Resources in our k8s cluster. The problem is that Azure managed Prometheus, doesn't pick up the PrometheusRules by default. We are therefore building a simple k8s controller to glue things together (i.e. to reconcile the Azure equivalent resources with what is configured on the PrometheusRule Custom Resource).

moshemal:
> Hey @moshemal 👋, Thanks for looking into this.
> 
> I guess it might work since YAML is a superset of JSON but that's not exactly my main problem. The tool currently tries to read the input from a file and that's what I would like to avoid. I would like to "pipe" a JSON string directly into the tool instead.
> 
> For a bit more context, we depend on a tool (a k8s controller) that generates [PrometheusRule](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PrometheusRule) Custom Resources in our k8s cluster. The problem is that Azure managed Prometheus, doesn't pick up the PrometheusRules by default. We are therefore building a simple k8s controller to glue things together (i.e. to reconcile the Azure equivalent resources with what is configured on the PrometheusRule Custom Resource).

I am not sure that your PR solve the stream issue. Did you check it with piping a string content to the command?
I think the stream should be read from stdin

monteiro-renato:
Much appreciated!

PR #1007, opened 2024-11-04
Adding UX recording rules. Enabling Azure portal insight blade

ehrnst:
# PR Description
This PR will enable the UX recording rules to allow the `insights` blade to work for the AKS clusters. The only change is to the addon template.

image

# New Feature Checklist

i have not checked any of the boxes, as i dont see how they fit.


bragi92:
Hi @ehrnst Thanks for making this PR. This feature (UX recording rules for managed prometheus) is still in preview and we can't enable it by default as of right now. We will need to add this in all the templates (ARM, Bicep, Policy, terraform) and will be making the change when it goes out of preview and hence can't merge this in yet. Does this sound good?

PR #1144, opened 2025-05-07
fix: move filesystemwatcher to hash based golang appoach for windows

bragi92:
# PR Description

Test Image : 6.16.0-kaveesh-filesystemwatcher-fix-05-08-2025-f8c65f5d

This PR replaces the PowerShell-based FileSystemWatcher script with a new implementation in Golang (shared.CheckForFilesystemChanges) that uses SHA256 hashing to detect changes in monitored configuration files.

Key Changes
✅ New Go Function: Added CheckForFilesystemChanges() in the shared package.

🔁 Behavior:

Computes a SHA256 hash over contents and paths of all files in:

`C:\etc\config\settings`

`C:\etc\config\settings\prometheus`

`TokenConfig.json` (conditionally, if present)

Compares it to the last stored hash (`C:\last_config_hash.txt`)

If the hash has changed:

Logs a message with a timestamp to `C:\filesystemwatcher.txt`

Updates the stored hash

🧹 Eliminates Dependence on PowerShell:

Removes the need for FileSystemWatcher and event subscription cleanup.

Avoids keeping a PowerShell process running persistently.

Rationale
🪶 Lower Overhead: For a small number of files checked every minute, a hashing approach is simpler and resource-light.

📦 Container Compatibility: Avoids potential complications with long-running PowerShell loops in Windows containers.

🔐 Deterministic and Explicit: Changes are only logged when actual content or file names differ.

Performance Considerations
Only ~10 small files checked per minute.

CPU/memory impact is negligible due to the small size and infrequent polling.

Suitable for running inside resource-constrained Windows containers.

# New Feature Checklist

[comment]: # (The below checklist is for code changes. Not all boxes necessarily need to be checked. Build, doc, and template changes do not need to fill out the checklist.)
# Tests Checklist

- [ ] Have end-to-end Ginkgo tests been run on your cluster and passed? To bootstrap your cluster to run the tests, follow [these instructions](/otelcollector/test/README.md#bootstrap-a-dev-cluster-to-run-ginkgo-tests).
    - [ ] `operator`
    - [ ] `windows`
    - [ ] `arm64`
    - [ ] `arc-extension`
    - [ ] `fips`
- [ ] Have new tests been added? For features, have tests been added for this feature? For fixes, is there a test that could have caught this issue and could validate that the fix works?
  - [ ] Is a new scrape job needed?
    - [ ] The scrape job was added to the folder [test-cluster-yamls](/otelcollector/test/test-cluster-yamls/) in the correct configmap or as a CR. 
  - [ ] Was a new test label added?
    - [ ] A string constant for the label was added to [constants.go](/otelcollector/test/utils/constants.go).
    - [ ] The label and description was added to the [test README](/otelcollector/test/README.md).
    - [ ] The label was added to this [PR checklist](/.github/pull_request_template).
    - [ ] The label was added as needed to [testkube-test-crs.yaml](/otelcollector/test/testkube/testkube-test-crs.yaml).
  - [ ] Are additional API server permissions needed for the new tests?
    - [ ] These permissions have been added to [api-server-permissions.yaml](/otelcollector/test/testkube/api-server-permissions.yaml).
  - [ ] Was a new test suite (a new folder under `/tests`) added?
    - [ ] The new test suite is included in [testkube-test-crs.yaml](/otelcollector/test/testkube/testkube-test-crs.yaml).


vishiy:
Hi Kaveesh - so the late dir creation (after starting watcher) is the root cause ?

gracewehner:
Thanks Kaveesh we can probably add a liveness probe test similar to [what we have](https://github.com/Azure/prometheus-collector/blob/main/otelcollector/test/ginkgo-e2e/livenessprobe/liveness_test.go#L81) for checking if a process is not running or if the configmap changed, but for running a command in the container to change the tokenconfig.json file or that directory and checking if the pods restart after

bragi92:
> Hi Kaveesh - so the late dir creation (after starting watcher) is the root cause ?

Hey @vishiy : no I think the root cause is something else (like flaky windows event change communication and one of the checks might have been doing an incorrect thing in powershell. That initial commit was just to add some debug logs and my hunch was that it was because of late dir creation but that appears to be untrue. Currently moving everything to a hash based approach.



> Thanks Kaveesh we can probably add a liveness probe test similar to [what we have](https://github.com/Azure/prometheus-collector/blob/main/otelcollector/test/ginkgo-e2e/livenessprobe/liveness_test.go#L81) for checking if a process is not running or if the configmap changed, but for running a command in the container to change the tokenconfig.json file or that directory and checking if the pods restart after

Makes sense @gracewehner , added a test for it.

otelcollector/main/main.go
```diff
@@ -51,9 +51,6 @@ func main() {
 				log.Fatal(err)
 			}
 		}
-	} else if osType == "windows" {
-		fmt.Println("Starting filesystemwatcher.ps1")
-		shared.StartCommand("powershell", "-NoProfile", "-ExecutionPolicy", "Bypass", "-File", "C:\\opt\\scripts\\filesystemwatcher.ps1")
```
bragi92, line 56: calculate hash once during startup so that it takes any change that happens in the first 5 mins.

otelcollector/main/main.go
```diff
@@ -350,7 +347,8 @@ func healthHandler(w http.ResponseWriter, r *http.Request) {
 			goto response
 		}
 	} else {
-		if shared.HasConfigChanged("C:\\opt\\microsoft\\scripts\\filesystemwatcher.txt") {
+		shared.CheckForFilesystemChanges()
```
bragi92, line 350: if file exists, no need to check the hash even

otelcollector/shared/filesystemwatcher_utility.go
```diff
@@ -0,0 +1,90 @@
...
+	// Generate final combined hash
+	finalHash := hex.EncodeToString(h.Sum(nil))
+
+	// Compare to last stored hash
+	lastHashBytes, _ := os.ReadFile(hashStore)
+	lastHash := string(lastHashBytes)
+	lastHash = strings.TrimSpace(lastHash)
+
+	// Write the new hash on the first run, but don't log yet since it's the first time
+	if lastHash == "" {
```
bragi92, line 56: remove the first hash check here, do it in main.go during initial start up

rashmichandrashekar (approved):

vishiy (approved):

PR #1298, opened 2025-09-10
chart: Match AKS RP chart for values and daemonset yaml

gracewehner:
- Chart needs to match AKS RP changes for the extension migration bug bash
- Both ports are there currently since those were the initial changes. I will remove grpc port for public preview with next addon release until we have in the API that that port is configurable
- The structure of the adapter chart for the values.yaml had been changed too, this now matches that

### DIFF
--- a/otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-daemonset.yaml
+++ b/otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-daemonset.yaml
@@ -134,11 +134,22 @@ spec:
             - name: MINIMAL_INGESTION_PROFILE
               value: "true" # only supported value is the string "true"
             - name: APPMONITORING_AUTOINSTRUMENTATION_ENABLED
-              value: "{{ .Values.AzureMonitorMetrics.IsAppMonitoringAutoInstrumentationEnabled }}"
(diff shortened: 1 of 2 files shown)


rashmichandrashekar (approved):

otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-daemonset.yaml
```diff
@@ -414,11 +425,22 @@ spec:
...
             - name: APPMONITORING_OPENTELEMETRYMETRICS_PORT
-              value: "{{ .Values.AzureMonitorMetrics.OpenTelemetryMetricsPort }}"
+              value: "{{ .Values.AppmonitoringAgent.openTelemetryMetricsPort | default 28333 }}"
+          {{- if .Values.AppmonitoringAgent.isOpenTelemetryMetricsEnabled | default false }}
+          ports:
+            - name: otlp-http-port
+              containerPort: 56681
+              hostPort: {{ .Values.AppmonitoringAgent.openTelemetryMetricsPort | default 28333 }}
+              protocol: TCP
+            - name: otlp-grpc-port
```
vishiy, line 439: need grpc now ? may be add this when grpc is fully supported ? this way, we we dont fail if this specific port is taken though its not used ?

rashmichandrashekar (approved):

PR #1312, opened 2025-09-29
fix: Values context for Arc chart

gracewehner:
- The new variables added at the top of daemonset.yaml for the t-shirt sizing causes a context change for the template values. Explicitly use $.Values. instead of .Values to point to the root.
- This was causing an issue installing the Arc extension where it was erroring with ".Values.Azure" not found

### DIFF
--- a/otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-daemonset.yaml
+++ b/otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-daemonset.yaml
@@ -100,7 +100,7 @@ spec:
             - name: customGlobalEndpoint
               value: {{ required "customGlobalEndpoint is required in Arc Autonomous" $.Values.AzureMonitorMetrics.arcAutonomousSettings.customGlobalEndpoint | toString | trim | quote }}
             - name: customResourceEndpoint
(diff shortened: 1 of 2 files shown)


otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-daemonset.yaml
```diff
@@ -547,7 +547,7 @@ spec:
            - --token-server-listening-port=7777
            - --health-server-listening-port=9999
            - --restart-pod-waiting-minutes-on-broken-connection=240
-          image: "{{ .Values.AzureMonitorMetrics.AddonTokenAdapter.ImageRegistry }}{{ .Values.AzureMonitorMetrics.AddonTokenAdapter.ImageRepositoryWin }}:{{ .Values.AzureMonitorMetrics.AddonTokenAdapter.ImageTagWin }}"
+          image: "{{ $.Values.AzureMonitorMetrics.AddonTokenAdapter.ImageRegistry }}{{ $.Values.AzureMonitorMetrics.AddonTokenAdapter.ImageRepositoryWin }}:{{ $.Values.AzureMonitorMetrics.AddonTokenAdapter.ImageTagWin }}"
```
rashmichandrashekar, line 550: Hi @gracewehner - Could you pls double check this renders correctly by doing a helm template for the aks daemonset chart too (bascically double check this works as expected outside the loop)?

gracewehner: Thanks Rashmi, yes did this and everything gets substituted as expected

PR #1482, opened 2026-03-27
build: fix retina images coming from ghcr. Use repo helm chart and replicate to mcr

gracewehner:
Using suggestion from retina team to checkout the repo and use the chart from there
Uses the Permissive network policy that's suggested in the 1ES docs for github and helm access


.trivyignore
```diff
@@ -3,5 +3,7 @@
 # CRITICAL
 CVE-2026-33186 #google.golang.org/grpc
 # HIGH
-
-# MEDIUM
```
rashmichandrashekar, line 7: this is updated, pls fix

PR #1582, opened 2026-06-23
fix: don't drop entire v2 config when an optional section file is missing

bragi92:
# PR Description

## Problem

A **partial v2 `ama-metrics-settings-configmap`** that omits the `prometheus-collector-settings` section causes the **entire v2 configuration to be silently discarded**, falling back to defaults.

This surfaced as two separate customer escalations:

- **Data-plane:** a custom `kubestate` keep-list is ignored, so PV/PVC kube-state metrics — `kube_persistentvolumeclaim_info`, `kube_persistentvolume_capacity_bytes`, `kube_pod_spec_volumes_persistentvolumeclaims_info` — are never ingested. `kube_pod_labels` still appears (it matches `kube_.*_labels` in the minimal-ingestion baseline), which masked the bug and made it look intermittent.
- **Control-plane (CCP):** the whole `controlplane-metrics` section is discarded, so `kube-scheduler` / `kube-controller-manager` scrape jobs are never created even when set to `true`; only the default-on `apiserver`/`etcd` work.

`schema-version: v1` is unaffected.

## Root cause

`shared.ParseMetricsFiles` opens each file in a hard-coded list and returns `nil` for the **whole** config on the first `os.Open` error. Both callers pass a 2-file list whose second entry is `prometheus-collector-settings`:

| Caller | File list |
|---|---|
| data-plane `mp/configmapparser.go` | `[cluster-metrics, prometheus-collector-settings]` |
| control-plane `ccp/configmapparserforccp.go` | `[controlplane-metrics, prometheus-collector-settings]` |

Each configmap section is projected by Kubernetes as its own file under `/etc/config/settings`. A partial configmap that omits `prometheus-collector-settings` has no such file → `os.Open` fails with `NotExist` → the correctly-parsed `cluster-metrics`/`controlplane-metrics` map is thrown away. The caller logs `Using defaults as error parsing files: ...` and continues with a `nil` map, so every downstream section parser uses defaults.

`v1` uses `ParseV1Config`, which reads the settings *directory* and treats each file independently, so a missing file is never fatal — hence the v1/v2 discrepancy.

Confirmed from a customer pod log:
```
AZMON_AGENT_CFG_SCHEMA_VERSION=v2
Using defaults as error parsing files: failed to open file /etc/config/settings/prometheus-collector-settings: no such file or directory
```

## Fix

Treat a **missing file as an omitted optional section**: skip it (`os.IsNotExist`) and continue parsing the rest. Downstream section parsers already guard every lookup with `if section, ok := metricsConfigBySection[...]; ok`, so absent sections correctly fall back to defaults. Only **non-`NotExist` open errors** (e.g. permissions) and **read errors** remain fatal. The skip is logged with the section name, e.g. `optional settings section "prometheus-collector-settings" not provided ...; using defaults`.

Per-file parsing is extracted into `parseMetricsFileInto`, which also closes each handle promptly (fixing a pre-existing `defer`-in-loop fd leak).

This single change fixes both the data-plane and control-plane paths, since both call `ParseMetricsFiles`.

## Tests

All new tests **fail without the fix** and pass with it (verified by reverting `helpers.go`):

- `shared/helpers_test.go`
  - `TestParseMetricsFiles_MissingOptionalFileIsSkipped` — data-plane: `cluster-metrics` keep-list survives a missing `prometheus-collector-settings`.
  - `TestParseMetricsFiles_ControlplanePartialConfigmap` — control-plane: `kube-scheduler = true` survives.
  - `TestParseMetricsFiles_AllFilesPresent`, `TestParseMetricsFiles_AllMissingReturnsEmpty`.
- `shared/configmap/ccp/parsemetricsfiles_partial_test.go`
  - `TestParseControlplaneScrapeSettings_PartialConfigmap` — end-to-end through the CCP consumer: asserts `controlplane-kube-scheduler/-kube-controller-manager/-etcd/-apiserver` are enabled from a partial v2 configmap.

### Verification (local)

- `go test ./...` (module `shared`) — all 4 new `TestParseMetricsFiles_*` tests pass.
- `go test ./...` (module `ccp`) — passes, including the new CCP regression test.
- `go build ./...` — clean for `shared`, `mp`, and `ccp`. `gofmt` clean.
- Negative control: reverting the fix makes all three new tests fail with the original `failed to open file ... prometheus-collector-settings` error.

> Two unrelated tests (`TestTailMELogFile`, `TestTailMELogFile_NoNewContent` in `me_log_tailer_test.go`) fail **on Windows only** during `t.TempDir()` cleanup (the log tailer holds the file open; Windows can't unlink an open file). Confirmed they fail identically on pristine `origin/main`, so they are pre-existing and independent of this change.

> Note: the `mp` unit-test package does not currently compile on `main` (pre-existing: `tomlparser_debug_mode_test.go` and `tomlparser-pod-annotation-based-scraping_test.go` target an obsolete file-based API). Repairing those is out of scope here, so the data-plane regression is covered at the `shared` level — the exact function that was buggy and that the `mp` path calls. These unit packages are not compiled by CI today (config tests run via testkube e2e).

## Customer mitigation (until the new image ships)

Add a `prometheus-collector-settings` section to the v2 configmap (any content, e.g. `cluster_alias = ""`), or use `schema-version: v1`. The canonical sample configmap already includes this section, so users who start from it are unaffected.

## Repro evidence (engineer cluster, image `7.0.0-main-05-07-2026`)

In-pod `KUBESTATE_METRICS_KEEP_LIST_REGEX` + workspace `count()`:

| Config | the 3 PV/PVC metrics |
|---|---|
| v1 | collected |
| v2 **without** `prometheus-collector-settings` | **dropped** |
| v2 **with** `prometheus-collector-settings` | collected |


bragi92:
### Note for reviewers: the failing **Build: run trivy scan** checks are pre-existing and unrelated to this PR

This PR only changes `otelcollector/shared/helpers.go` and adds test files — no dependency, base-image, or `.trivyignore` changes. The Trivy failures come from newly-published CVEs against currently-pinned third-party libs and the Azure Linux base image, and they affect `main` directly.

Evidence this is not from this change:
- `.trivyignore` on this branch is **identical to `main`** (untouched here).
- `main`'s own `scan-last-released-image` workflow is currently **failing** with the same condition (no PR involved).
- None of the failing CVEs are suppressed in `main`'s `.trivyignore` — so any branch cut from `main` right now fails Trivy identically.

Failing set (across the prom-collector, CCP, config-reader, and target-allocator image scans):

| Library | Type | Installed → Fixed | Example CVEs |
|---|---|---|---|
| openssl / openssl-libs | os-pkg (base image) | `3.3.5-5.azl3` → `3.3.7-2.azl3` | CVE-2026-34182 (CRITICAL) |
| golang.org/x/net | gobinary | `v0.48.0` → `0.55.0` | 25680, 25681, 27136, 39821, 42502, 42506 |
| golang.org/x/crypto (ssh) | gobinary | `v0.49.0` → `0.52.0` | 39827–39835, 42508, 46595/97/98 |
| stdlib (Go toolchain) | gobinary | `v1.25.9` → `1.25.11 / 1.26.4` | 27145, 42504, 42507 |

These should be addressed by the routine `.trivyignore` / dependency-bump maintenance on `main` (ideally upgrading `x/net`, `x/crypto`, and the Go toolchain, and suppressing the openssl CRITICAL until the next Azure Linux base-image rebuild, consistent with the existing glibc entries). Intentionally **not** bundling that security maintenance into this parser fix to keep the change focused; this PR should not be blocked on the Trivy gate.

rashmichandrashekar:
Thanks Kaveesh. whats the behavior today with v1? will these changes have impact on the v1 processing as well?

rashmichandrashekar:
@bragi92 - should we also add tests to make sure the full configmap parsing works fine after this for v1 and v1?

bragi92:
> Thanks Kaveesh. whats the behavior today with v1? will these changes have impact on the v1 processing as well?

@rashmichandrashekar V1 is unaffected. It uses a different function for parsing the configmap and converting it into the common data structure so we should be good here.

bragi92:
> @bragi92 - should we also add tests to make sure the full configmap parsing works fine after this for v1 and v1?

@rashmichandrashekar I did add unit tests for this scenario but its not actually called in the build pipeline. The tests did succeed but they had to be run manually. Taking a look to see what it would take to add these to our build pipeline.

bragi92:
@rashmichandrashekar test added for partial configmap.

rashmichandrashekar (approved):

PR #1584, opened 2026-06-23
ci: disable linux python reference app build (pypi.org egress failure)

bragi92:
## Problem

Every `Azure.prometheus-collector` build has failed since ~2026-06-15. On otherwise-green PRs (e.g. #1582, #1583) the **only** remaining red leg is **`Build: linux python reference app image`**.

It fails in the Docker build at:

```
#10 [3/4] RUN pip install prometheus-client opentelemetry-api opentelemetry-sdk opentelemetry-exporter-otlp-proto-http
  WARNING: Retrying ... 'Connection to pypi.org timed out. (connect timeout=15)'   (x5)
  ERROR: Could not find a version that satisfies the requirement prometheus-client
ERROR: failed to solve: process "/bin/sh -c pip install ..." did not complete successfully: exit code: 1
```

The 1ES CI agents can no longer reach **pypi.org**, so the `pip install` in `internal/referenceapp/python/linux/Dockerfile` fails with connect timeouts (5 built-in retries exhausted). This is purely environmental — re-running does not help (multiple consecutive builds fail identically).

## Fix

Comment out the `Linux_Python_Reference_App` job in `.pipelines/azure-pipeline-build.yml`, **mirroring PR #1390**, which disabled the *Windows* python reference app for this exact pypi.org reason.

### Why this is safe
- The job is a **leaf** — no other job has `dependsOn: Linux_Python_Reference_App`.
- The reference-app test manifests under `internal/referenceapp/*.yaml` **pin previously-published image tags** (e.g. `…-ref-app-python`), not the per-PR build output, so no test/deploy stage consumes this job's artifact.
- The python reference app is single-arch (amd64), so no multi-arch manifest-combine job depends on it.

### Scope
- Touches only `.pipelines/azure-pipeline-build.yml` (the PR/CI build pipeline that gates merges), exactly as PR #1390 did. The official release pipeline (`OneBranch.Official.yml`) is intentionally left unchanged, consistent with that precedent.

Re-enable this job once egress to pypi.org (or an approved internal mirror) is restored.

## Validation
- `git diff` is scoped to the single job block (85 lines commented + 2-line reason comment).
- YAML still parses (`yaml.safe_load`).

### DIFF
--- a/.pipelines/azure-pipeline-build.yml
+++ b/.pipelines/azure-pipeline-build.yml
@@ -412,91 +412,93 @@ extends:
           workingDirectory: $(Build.ArtifactStagingDirectory)/refappgolanglinux/
           displayName: "ORAS Push Artifacts in $(Build.ArtifactStagingDirectory)/refappgolanglinux/"
           condition: succeeded()
-      - job: Linux_Python_Reference_App
-        displayName: "Build: linux python reference app image"
-        pool:
-          name: Azure-Pipelines-CI-Test-EO
-          image: ci-1es-managed-ubuntu-2204
-          os: linux
-        dependsOn: Image_Tags_and_Ev2_Artifacts
-        variables:
-          skipComponentGovernanceDetection: true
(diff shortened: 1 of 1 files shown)


bragi92:
Superseded by #1583. The build gate blocks merges and requires BOTH the Trivy fix and this python-ref-app disable in a single green PR, so I folded this exact change into #1583 (cherry-picked, identical hunk). Closing to avoid duplicate open PRs.

PR #1590, opened 2026-06-24
feat: support azureMonitorProfile.metrics.controlPlane.enabled in onboarding templates (ARM/Bicep/Terraform/Policy)

bragi92:
## What

Plumbs the new AKS API property **`azureMonitorProfile.metrics.controlPlane.enabled`** through the four onboarding IaC templates in this repo, so clusters can be opted in/out of Azure Monitor managed Prometheus **control-plane** metrics (default targets: `controlplane-apiserver`, `controlplane-etcd`) declaratively.

Pairs with the azure-cli change **Azure/azure-cli#33537** (`--enable/--disable-control-plane-metrics`).

## Changes

| Template | Change |
|---|---|
| **ARM** (`AddonArmTemplate`) | New `enableControlPlaneMetrics` parameter (default `false`) + `controlPlane.enabled` under `azureMonitorProfile.metrics`; bumped the managed-cluster PUT apiVersion `2023-01-01` → `2026-06-01`. Added to the parameters file. |
| **Bicep** (`AddonBicepTemplate`) | New `enableControlPlaneMetrics` param threaded into the nested profile module; `controlPlane.enabled` added; apiVersion bumped. Added to parameters file. |
| **Policy** (`AddonPolicyTemplate`) | New `enableControlPlaneMetrics` policy parameter, threaded through the deployment into `azureMonitorProfile.metrics.controlPlane.enabled`; apiVersion bumped. The `existenceCondition` is intentionally left checking only `metrics.enabled` (control-plane metrics is opt-in and should not drive compliance). |
| **Terraform** (`AddonTerraformTemplate`) | New `enable_control_plane_metrics` variable. The `azurerm` `monitor_metrics` block can't express this property, so it's set via the **`azapi`** provider (`azapi_update_resource`), added to `providers.tf`. |
| READMEs | Documented the new opt-in flag for all four. |

Default is **opt-in / `false`** everywhere, matching the CLI's opt-in flag and preserving current template behavior.

## Greenfield race handling

The CLI PR notes that setting `controlPlane.enabled=true` before the cluster's DCRA exists makes the RP schedule the control-plane-metrics (CCP) collection pod prematurely ΓåÆ CrashLoopBackOff until reconcile. These templates avoid that **by ordering**:

- ARM/Bicep/Policy: the managed-cluster profile PUT already `dependsOn` the DCRA deployment, so `controlPlane.enabled` is only applied after the DCRA exists.
- Terraform: the `azapi_update_resource` has `depends_on = [azurerm_monitor_data_collection_rule_association.dcra]`.

## API version

`azureMonitorProfile.metrics.controlPlane` went **GA in `2026-04-01`** — it is absent from every stable version through `2026-01-01`, first appeared in `2026-02-02-preview`, and has been in every stable version since. These templates use **`2026-06-01`, the latest stable AKS API version**, which still carries the property. The paired aks-preview CLI extension ([Azure/azure-cli-extensions#9931](https://github.com/Azure/azure-cli-extensions/pull/9931)) vendors `2026-04-02-preview`; for these customer-facing IaC templates the latest **GA stable** is preferred.

All four managed-cluster call sites use `2026-06-01` (ARM/Bicep/Policy resource `apiVersion`, Terraform `azapi` `type`). The Bicep CLI's bundled type catalog does not yet include `2026-06-01`, so the nested module suppresses the benign `BCP081` "types not available" warning; the compiled ARM correctly emits `controlPlane.enabled` and deploys fine.

### Other resource API versions

Audited every resource type in these templates against the registered provider versions and the published stable swagger:

| Resource type | Version | |
|---|---|---|
| `Microsoft.ContainerService/managedClusters` | `2026-06-01` | bumped — latest stable |
| `Microsoft.Insights/dataCollectionRuleAssociations` | `2025-05-11` | bumped — latest stable, now matches the DCR version |
| `Microsoft.Insights/dataCollectionRules` | `2025-05-11` | already latest stable |
| `Microsoft.Insights/dataCollectionEndpoints` | `2024-03-11` | already latest stable |
| `Microsoft.AlertsManagement/prometheusRuleGroups` | `2023-03-01` | already latest stable |
| `Microsoft.Authorization/roleAssignments` | `2022-04-01` | already latest stable |
| `Microsoft.Dashboard/grafana` | `2025-08-01` | left as-is — see below |

`Microsoft.Dashboard/grafana` has a newer stable (`2026-09-01`), but nothing in this PR needs it and Bicep has no types for it yet, so bumping would trade compile-time validation for no functional gain. Left at `2025-08-01`.

## Live deployment validation

Deployed the ARM template end to end against a real cluster (`eastus2`, AKS 1.35, 2 nodes) with
`enableControlPlaneMetrics=true`, alongside a new Azure Monitor Workspace and Managed Grafana.

**Result: `provisioningState: Succeeded`.** All resources created — DCE, DCR, DCRA, 6 recording
rule groups, Grafana→AMW integration, and the Monitoring Data Reader role assignment.

Cluster property, read back at `2026-06-01`:

```json
"azureMonitorProfile": {
  "metrics": {
    "enabled": true,
    "controlPlane": { "enabled": true },
    "kubeStateMetrics": { "metricLabelsAllowlist": "", "metricAnnotationsAllowList": "" }
  }
}
```

Agent pods, all healthy:

```
ama-metrics-797764c94f-6xpvv                    2/2  Running
ama-metrics-797764c94f-q5wmc                    2/2  Running
ama-metrics-ksm-d484f9df8-69g8w                 1/1  Running
ama-metrics-node-2xtgr                          2/2  Running
ama-metrics-node-6kj2z                          2/2  Running
ama-metrics-operator-targets-74f955d776-qh94f   2/2  Running
```

Scrape jobs reporting into the AMW — **both control plane jobs present**:

| Job | Targets |
|---|---|
| `controlplane-apiserver` | 2 |
| `controlplane-etcd` | 3 |
| cadvisor / kubelet / node | 2 each |
| kube-state-metrics | 1 |
| networkobservability-retina | 2 |

Control plane metric samples queried from the AMW Prometheus endpoint:

| Metric | Value |
|---|---|
| `apiserver_request_total` | 27,753 |
| `apiserver_request_duration_seconds_sum` | 495,383 |
| `apiserver_current_inflight_requests` | 4 |
| `etcd_server_has_leader` | 3 |
| `etcd_mvcc_db_total_size_in_bytes` | 24,604,672 |

Standard metrics also flowing: `kube_pod_info` 33, `node_cpu_seconds_total` 64 series,
`container_memory_working_set_bytes` ~12.9 GB, `kube_node_status_condition{Ready}` 2.

Worth noting: the `AzureMonitorMetricsControlPlanePreview` AFEC flag was **Unregistered** on the
test subscription, so this exercises the GA API-property path rather than the legacy feature-flag
gating.

## Bug found and fixed during validation

The first deployment attempt failed outright:

```
UnmarshalError: Invalid request body. Converting request body to a managed cluster
encountered error: json: unknown field "mode"
```

and after removing `mode`, the same error for `id`.

The nested managed-cluster resource in the ARM and Policy templates carried `"mode":
"Incremental"` and `"id"` inside its `properties`. Neither is a ManagedCluster property — `mode`
belongs on the enclosing deployment, and the cluster is already addressed by name plus the nested
deployment's subscription and resource group.

**These fields pre-date this PR.** The `2023-01-01` RP silently ignored them; from `2026-06-01`
the RP rejects unknown fields, so the API version bump turned a latent template bug into a hard
deployment break. Both are now removed. Bicep and Terraform never carried them.

## Static checks

- All edited JSON (ARM, ARM params, Policy rules, Policy params, Bicep params) parses.
- `az bicep build` succeeds (exit 0) for all three Bicep files; both nested modules build **warning-free**, and the compiled ARM contains `"apiVersion": "2026-06-01"` and `controlPlane.enabled`.
- `az deployment group validate` passes.
- Terraform: HCL added by hand (no binary in the authoring env); `terraform init -upgrade && terraform validate` recommended before merge.

Worth noting that **all of the above passed while the `mode`/`id` bug was still present.** JSON parsing, `bicep build` and ARM preflight validation do not catch invalid properties inside a nested managed-cluster body; only a real deployment surfaced it.

## Unrelated CI fix included

`otelcollector/build/linux/Dockerfile` copied `/usr/lib/libp11-kit.so.0.3.0` by its full version.
That file is the target of the `p11-kit-proxy.so` and `libp11-kit.so.0` symlinks so it has to be
copied, but its name changes whenever the base image is rebuilt.

`mcr.microsoft.com/azurelinux/base/core:3.0` is a floating tag and was rebuilt **2026-08-20** with
p11-kit `0.26.5-1.azl3`, which ships `libp11-kit.so.0.4.11`. Every linux AMD64/ARM64 image build
has failed since:

```
ERROR: failed to compute cache key: failed to calculate checksum of ref ...:
"/usr/lib/libp11-kit.so.0.3.0": not found
```

Reproduced against the current base image — only the pinned path is missing:

```
OK      /usr/lib/p11-kit-trust.so
OK      /usr/lib/p11-kit-proxy.so   -> libp11-kit.so.0.4.11
MISSING /usr/lib/libp11-kit.so.0.3.0
OK      /usr/lib/libnssckbi.so
OK      /usr/lib/libp11-kit.so.0    -> libp11-kit.so.0.4.11
```

Fixed by globbing `libp11-kit.so.0*`, which picks up the SONAME symlink and whichever versioned
file it points to. Verified by building the same base-to-distroless COPY sequence locally: the
layer builds and `libp11-kit.so.0.4.11` lands in the image with both symlinks resolving.

This is not related to the template changes, but it blocked CI on this branch and every other open
PR (an unrelated dependabot build failed identically an hour before this one), so it is fixed here.

### Same class of bug, also unpinned

While confirming the root cause I checked every other `COPY --from=builder` path in both the main and CCP Dockerfiles. Three more were pinned to a full version:

| Library | Was | Now |
|---|---|---|
| `libboost_filesystem.so` | `1.83.0` | `libboost_filesystem.so.*` |
| `libboost_atomic.so` | `1.83.0` | `libboost_atomic.so.*` |
| `libcpprest.so` | `2.10` | `libcpprest.so.*` |

Unlike most libraries copied here, these have **no stable SONAME symlink** to point at instead: the fully qualified version *is* the SONAME, so there is no `libboost_filesystem.so.1` or `libcpprest.so.2` to reference. Verified by reproducing the builder stage locally, where only the versioned regular files exist and no bare `.so` symlinks are present.

They resolve today because azurelinux still ships `boost 1.83.0-2.azl3`, so nothing is broken right now, but it is the identical failure mode waiting on the next package refresh, in both Dockerfiles. Globbed for the same reason.

Every remaining `COPY --from=builder` path uses a stable major-version SONAME (`.so.6`, `.so.1`, `.so.0`, and `libsymcrypt.so.103` which is itself a symlink), so nothing else here is version-fragile.

## Remaining items for reviewer

1. **Terraform `azapi` version.** Pinned `~> 1.15` and used a `jsonencode` body (1.x style). If you prefer azapi 2.x, the `body` becomes an HCL object (drop `jsonencode`).
2. **Default value.** Defaulted to `false` (opt-in). Confirm that's the desired default for these onboarding templates vs. on-by-default.


AddonArmTemplate/FullAzureMonitorMetricsProfile.json
```diff
@@ -178,16 +185,17 @@
...
               "properties": {
-                "mode": "Incremental",
-                "id": "[parameters('clusterResourceId')]",
                 "azureMonitorProfile": {
                   "metrics": {
                     "enabled": true,
                     "kubeStateMetrics": {
                       "metricLabelsAllowlist": "[parameters('metricLabelsAllowlist')]",
                       "metricAnnotationsAllowList": "[parameters('metricAnnotationsAllowList')]"
+                    },
```
rashmichandrashekar, line 196: with the new api-version -if i put is made without this field, what happens?

rashmichandrashekar (approved):

AddonArmTemplate/FullAzureMonitorMetricsProfile.json
bragi92: verified on a test cluster. Property value is preserved if the field is omitted.

PR #1610, opened 2026-07-10
ci: fix size-label workflow for fork PRs

bragi92:
## Summary
- switch size-label workflow trigger from pull_request to pull_request_target
- grant issues:write permission for label API updates
- keep pull-requests:write and contents:read unchanged

## Why
Fork-based pull requests run with restricted token permissions under pull_request. The size-label action needs write access to apply labels and currently fails with 403 (Resource not accessible by integration).


fuweid (approved):
LGTM non-binding

rashmichandrashekar (approved):

PR #1651, opened 2026-07-29
[target-allocator] Wait for the config file instead of exiting when it is missing

boulajp:
## Problem
 
The Target Allocator reads its config file once at startup (`config.Load` -> `os.ReadFile`) and exits with a fatal error if the file isn't present yet.
 
This is fine when the config is mounted from a ConfigMap, but some deployments generate the config file into a volume shared with a sidecar container in the same Pod. Container start order within a Pod isn't guaranteed, so the Target Allocator can start before the sidecar has written the file. When that happens the process crash-loops until the sidecar wins the race. While the TA is down, collectors have no target assignments, which produces a scrape gap across the cluster on every restart or rollout where the race is lost.
 
## Fix
 
Add an opt-in `--config-file-wait-timeout` flag. When set to a positive duration, `Load` polls for the config file (every 250ms) until it exists and is non-empty, up to the timeout, before reading it. An empty file is treated as not-ready to guard against reading a partially written file.
 
The default is `0`, which preserves the existing fail-fast behavior, so this is a no-op for existing users unless they opt in.
 
The Azure Monitor metrics addon chart is also updated to set `--config-file-wait-timeout=5m` on the Target Allocator container, where the config file is written by the config-reader sidecar into a shared `emptyDir`.
 
# New Feature Checklist
 
- [ ] List telemetry added about the feature. _N/A - no new telemetry; only startup behavior changes, behind an opt-in flag._
- [ ] Link to the one-pager about the feature. _N/A - small reliability fix, no one-pager._
- [ ] List any tasks necessary for release (3P docs, AKS RP chart changes, etc.) after merging the PR. _The addon chart change (`--config-file-wait-timeout=5m`) is included in this PR; no other release tasks._
- [ ] Attach results of scale and perf testing. _N/A - the wait loop runs once at startup and is a no-op by default; no steady-state or scale impact._
 
# Tests Checklist
 
- [ ] Have end-to-end Ginkgo tests been run on your cluster and passed? _N/A - external contribution without cluster access; covered by the unit tests below._
    - [ ] `operator`
    - [ ] `windows`
    - [ ] `arm64`
    - [ ] `arc-extension`
    - [ ] `fips`
- [x] Have new tests been added? Added unit tests in `otel-allocator/internal/config/config_wait_test.go` covering: file already present, disabled timeout, file written after startup, empty-file timeout, never-appears timeout, the directory error case, an end-to-end `Load` test that succeeds when the config file is written late, and one confirming default fail-fast behavior.
  - [ ] Is a new scrape job needed? _No._
  - [ ] Was a new test label added? _No._
  - [ ] Are additional API server permissions needed for the new tests? _No._
  - [ ] Was a new test suite (a new folder under `/tests`) added? _No._

### DIFF
--- a/otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-targetallocator.yaml
+++ b/otelcollector/deploy/addon-chart/azure-monitor-metrics-addon/templates/ama-metrics-targetallocator.yaml
@@ -40,6 +40,7 @@ spec:
       - name: targetallocator
         args:
         - --enable-prometheus-cr-watcher
+        - --config-file-wait-timeout=5m
         image: "{{ .Values.AzureMonitorMetrics.ImageRegistry }}{{ .Values.AzureMonitorMetrics.ImageRepository }}:{{ .Values.AzureMonitorMetrics.ImageTagTargetAllocator }}"
         imagePullPolicy: IfNotPresent
         resources:
--- a/otelcollector/otel-allocator/internal/config/config.go
+++ b/otelcollector/otel-allocator/internal/config/config.go
@@ -45,6 +45,8 @@ const (
 	DefaultAllocationStrategy                          = "consistent-hashing"
 	DefaultFilterStrategy                              = "relabel-config"
 	DefaultCollectorNotReadyGracePeriod                = 30 * time.Second
+	DefaultConfigFileWaitTimeout                       = 0 * time.Second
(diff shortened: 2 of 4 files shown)


rashmichandrashekar:
@boulajp - thanks for making this pr. This repository only has a copy of the files in opentelemetry-operator - could you pls make the change in the OSS repo and it will automatically get picked up here? https://github.com/open-telemetry/opentelemetry-operator/tree/main/cmd/otel-allocator

PR #1690, opened 2026-08-24
Set environment variables directly instead of via a shell

bragi92:
## Summary

Removes a leftover from the shell-to-Go port: the environment-variable helper still appended `export KEY=value` to `~/.bashrc` and then ran `bash -c "source ~/.bashrc"` on every call (~100 call sites). Neither half does anything today, so this drops the shell entirely and adds input validation to the helper. Also fixes a latent bug that made `INVALID_CONFIG_FATAL_ERROR` unreadable on Linux.

## Why the `.bashrc` mechanism existed

Before `516c07e8` ("use golang for config processing") the agent was shell. `main.sh` was the ENTRYPOINT and invoked `configmap-parser.sh` as a subprocess. A child shell cannot mutate its parent's environment, so the parser wrote `export …` lines to `~/.bashrc` and the parent re-read them — `~/.bashrc` was the hand-off point between the two scripts, and the `config_*_env_var` files were written in `KEY=value` form specifically so they could be `source`d.

The Go port carried the pattern over near line-for-line, but it is no longer load-bearing:

- `os.Setenv` on the first line of the helper already sets the variable for this process.
- Every launcher in this package sets `cmd.Env = append(os.Environ())`, so `mdsd`, `MetricsExtension`, `otelcollector` and `promconfigvalidator` inherit from there.
- `exec.Command("bash", "-c", "source …")` ran in a **child** bash that exited immediately, so it could not affect this process's environment.
- **CCP already skipped the whole block** (`return if in ccp mode as no bash shell present here`) and behaves correctly in production — same parser code path.
- Nothing reads `~/.bashrc` any more. No `.sh` or Dockerfile sources it, and the liveness probe is now `httpGet /health:8080` rather than the old `livenessprobe.sh`.

The only observable difference is that `kubectl exec` into the pod followed by an interactive `bash` no longer pre-populates the agent's variables. `EchoVar` still logs them and `/opt/envvars.env` is still written.

## Changes

**`shared/process_utilities_linux.go`**
- `SetEnvAndSourceBashrcOrPowershell` now calls `os.Setenv` + `EchoVar` only; the `.bashrc` write and `source` are gone, along with the now-dead `CCP_METRICS_ENABLED` early return.
- Names must match `^[A-Za-z_][A-Za-z0-9_]*$`.
- Values may not contain `\n`, `\r` or NUL. The `configmapparser` env files are written and parsed line by line, so a line break in a value would be read back as an **additional** `key=value` pair and set a variable the caller never intended.
- The error log no longer prints `value`. Callers pass `echo=false` for secrets such as `APPLICATIONINSIGHTS_AUTH` to keep them out of the logs, but the error branch printed them regardless.
- Drops the now-unused `path/filepath` import.

**`shared/process_utilities_windows.go`** — same name/value validation before the `setx` path.

**`prom-config-validator-builder/main.go`** — fixes a real bug. The Linux branch wrote `export INVALID_CONFIG_FATAL_ERROR="…"`, but this file is parsed as `KEY=value` by `SetEnvVarsFromFile`, not sourced. The key therefore became the literal `export INVALID_CONFIG_FATAL_ERROR` (with the space), so `os.Getenv("INVALID_CONFIG_FATAL_ERROR")` in `fluent-bit/src/prometheus_collector_health.go:128` **always returned empty on Linux** and invalid-config errors never reached health telemetry. Now writes the same unquoted form the Windows branch already used. Control-character stripping widened from `\n` to `\r` and NUL as well.

**`shared/configmap/mp/configmapparser.go`** — `schema-version` and `config-version` are validated against `^[A-Za-z0-9._-]{1,10}$` and fall back to the defaults. Previously only spaces were stripped and the value truncated to 10 characters.

**`shared/configmap/mp/tomlparser-prometheus-collector-settings.go`** — `default_metric_account_name` must match `^[A-Za-z0-9][A-Za-z0-9._-]{0,63}$`; each `secrets_access_namespaces` entry must be a valid RFC 1123 label (≤63 chars). Invalid entries are logged and skipped. `cluster_alias` was already sanitized.

**`shared/configmap/mp/tomlparser-pod-annotation-based-scraping.go`** — the namespace regex is written unquoted. The `'…'` wrapper only mattered when the file was `source`d; the Go reader passes the quotes through as part of the value. Also removes the commented-out `linuxPrefix` block.

**`shared/file_utilities.go`, `shared/configmap/mp/configmapparser.go`** — `strings.SplitN(line, "=", 2)` instead of `strings.Split(...)` with a `len == 2` guard, so values containing `=` are no longer silently dropped. This matches what `handlePodAnnotationsFile` already did with `strings.Index`.

## Tests

New `shared/process_utilities_env_test.go`:
- normal values (including regexes, spaces, and values containing `=`) still reach `os.Getenv`
- malformed names are rejected, including the `export KEY` shape the validator file used to produce
- values with `\n` / `\r` / NUL are rejected, and no second variable is ever created from one value
- `~/.bashrc` is not created

```
ok  github.com/prometheus-collector/shared    0.015s
```

Builds verified for `GOOS=linux` and `GOOS=windows` across `shared`, `shared/configmap/mp`, `shared/configmap/ccp` and `prom-config-validator-builder`.

Note for reviewers: `shared/configmap/mp`'s test package does not compile on `main` today (`configurePodAnnotationSettings` / `ConfigureDebugModeSettings` are called with no arguments in the `_test.go` files). I confirmed this against a clean `origin/main` worktree — it is pre-existing and unrelated to this PR. `go vet ./...` on `shared` likewise reports pre-existing findings on lines this PR does not touch.

## Validation still to do

E2E on a live AKS cluster: confirm settings from the configmap continue to flow through to the collector config and that metrics land in the Azure Monitor workspace, plus the `INVALID_CONFIG_FATAL_ERROR` path now surfacing correctly. Keeping this as a draft until that passes.


otelcollector/shared/process_utilities_env_test.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, key := range keys {
+		t.Run(key, func(t *testing.T) {
+			if err := SetEnvAndSourceBashrcOrPowershell(key, "value", false); err == nil {
+				t.Errorf("expected error for key %q, got nil", key)
+			}
+			if key != "" && os.Getenv(key) != "" {
+				t.Errorf("key %q should not have been set", key)
+			}
+		})
+	}
```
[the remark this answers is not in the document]
bragi92: Fixed the determinism point in `0cebabd2` — switched to table-driven subtests with explicit names, and replaced `os.Getenv` with `os.LookupEnv`. `LookupEnv` distinguishes "set to empty" from "not set", so the empty-key case is now actually asserted rather than skipped by the old `if key != ""` guard, and a name that already exists in the environment is detected and skipped instead of producing a spurious failure.

One correction though: **the empty subtest name does not panic.** Go's testing package rewrites an empty or duplicate subtest name, so `t.Run("")` runs as `#00`. This test was passing before the change:

```
--- PASS: TestSetEnvAndSourceBashrcOrPowershell_RejectsInvalidKeys (0.01s)
    --- PASS: TestSetEnvAndSourceBashrcOrPowershell_RejectsInvalidKeys/#00 (0.01s)
    --- PASS: TestSetEnvAndSourceBashrcOrPowershell_RejectsInvalidKeys/export_AZMON_TEST (0.00s)
    ...
```

The named subtests are still a readability improvement, so the change stands — but flagging it so a reviewer doesn't read "will panic at runtime" as an unaddressed defect.

otelcollector/prom-config-validator-builder/main.go
```diff
@@ -38,19 +38,20 @@ func setFatalErrorMessageAsEnvVar(message string) {
...
-	re := regexp.MustCompile("\\n")
+	// Replace control characters so the value stays on a single line for env var parsing
+	re := regexp.MustCompile(`[\r\n\x00]`)
 	truncatedMessage = re.ReplaceAllString(truncatedMessage, "")
 
 	// Write env var to a file so it can be used by other processes
 	file, err := os.OpenFile("/opt/microsoft/prom_config_validator_env_var", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
 	if err != nil {
 		log.Println("prom-config-validator::Unable to open file - prom_config_validator_env_var")
 	}
```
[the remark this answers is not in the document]
bragi92: Took the error-handling fix in `0cebabd2` — early return when `OpenFile` fails, `defer file.Close()`, and both log lines now include the underlying error (previously they reported that something failed without saying what, which is unhelpful mid-incident).

Two clarifications on the diagnosis though:

**1. It does not panic.** `os.File` methods are nil-receiver-safe via `checkValid`, so they return `ErrInvalid` rather than dereferencing. Verified:

```
nil file Write  -> n=0 err=invalid argument (no panic)
nil file Close  -> err=invalid argument (no panic)
```

The real defect was quieter: on open failure the function carried on and the write silently could not succeed, with no indication why. That is worth fixing, just not for the stated reason.

**2. This is pre-existing code, not introduced by this PR.** The `OpenFile` block and the `Write` call are unchanged context lines either side of the one line this PR actually touched (the `setEnvVarString` format). Fixing it here anyway since the file is already open in this change and it is only a few lines.

rashmichandrashekar:
@bragi92 - could you pls run the configprocessing tests on this branch to make sure they work as expectd?

otelcollector/shared/configmap/mp/configmapparser.go
```diff
@@ -15,6 +16,10 @@ const (
 	defaultConfigFileVersion   = "ver1"
 )
 
+// configVersionPattern constrains the schema-version and config-version configmap values,
+// which are otherwise free-form text mounted from the settings configmap.
```
rashmichandrashekar, line 20: is this change relevant to this pr?

otelcollector/shared/configmap/mp/tomlparser-prometheus-collector-settings.go
```diff
@@ -8,12 +8,24 @@ import (
 	"strings"
 )
 
+// metricAccountNamePattern constrains default_metric_account_name to the characters a
+// Geneva metric account can actually contain.
+var metricAccountNamePattern = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9._-]{0,63}$`)
```
rashmichandrashekar, line 13: this is also not relevant to this pr i think - pls add this in the description and should be fine

otelcollector/shared/configmap/mp/configmapparser.go
bragi92: These two values are passed straight into  SetEnvAndSourceBashrcOrPowershell  (lines 26/32/44 and 50/56/68) — the helper this PR rewrites. It now rejects values containing  \n / \r /NUL, so without validating upstream these would start hitting the new error path and silently keep the old value instead of the documented default. The regex keeps the fallback behavior intact.

otelcollector/shared/configmap/mp/tomlparser-prometheus-collector-settings.go
bragi92: I needed to make this update too. The icm specifically had called out default_metric_account_name. Will update the description.

bragi92:
> bragi92 - could you pls run the configprocessing tests on this branch to make sure they work as expectd?

Yes, running. Will post back the results here.

rashmichandrashekar (approved):

